PR #2, opened 2022-08-18
Devices API rest endpoint migration

JamesReate:
- new structure
- rest apis
- tests


.idea/device-definitions-api.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
JamesReate, line 1: let's remove and gitignore the .idea folder

PR #3, opened 2022-08-22
Device definition grpc

rhvivancoeffio:


JamesReate:
Let's move the `server.go` and `service.go` in the `/internal/api/grpc` folder out of the grpc package just to `/internal/api`, ok to rename files to have grpc name in them, eg. `GrpcService` same as the type in the file. 

Linter is throwing some issues - I updated the readme on how to run it locally. I have resolved some of them, but there are others pending to be reviewed. There is also some tests failing, probably something needs to be updated.

PR #7, opened 2022-09-02
Device definition grpc update data

rhvivancoeffio:
# Proposed Changes
- Create new device definition endpoint
- Updated GRPC properties
 
### Impacted Routes

- UpdateDeviceDefinition


internal/core/commands/update_dd.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if err != nil {
+		return nil, &common.NotFoundError{
+			Err: fmt.Errorf("could not find device definition id: %s", command.DeviceDefinitionID),
+		}
+	}
+
+	deviceVehicleInfoMetaData := new(UpdateDeviceVehicleInfo)
+	if err := dd.Metadata.Unmarshal(deviceVehicleInfoMetaData); err == nil {
+		deviceVehicleInfoMetaData.FuelType = command.FuelType
```
JamesReate, line 85: maybe not something to address right now but what happens if we only want to set certain properties. eg. difference btw nil and empty. it may get complicated b/c we'd have to make them all pointers. Or we just make the contract explicit that everything gets updated.

JamesReate (approved):
some things to think about for later but good to merge now to keep moving

PR #11, opened 2022-09-12
Device definition update image

rhvivancoeffio:
# Proposed Changes
Add new GRPC Endpoint 

### Impacted Routes

SetDeviceDefinitionImage


internal/core/commands/update_dd_image.go
```diff
@@ -0,0 +1,76 @@
...
+func NewUpdateDeviceDefinitionImageCommandHandler(dbs func() *db.ReaderWriter, cache services.DeviceDefinitionCacheService) UpdateDeviceDefinitionImageCommandHandler {
+	return UpdateDeviceDefinitionImageCommandHandler{DBS: dbs, DDCache: cache}
+}
+
+func (ch UpdateDeviceDefinitionImageCommandHandler) Handle(ctx context.Context, query mediator.Message) (interface{}, error) {
+
+	command := query.(*UpdateDeviceDefinitionImageCommand)
+
+	dd, err := models.DeviceDefinitions(
+		qm.Where("id = ?", command.DeviceDefinitionID),
```
JamesReate, line 44: can also do models.DeviceDefinitionWhere.Id.EQ(command.DeviceDefinitionID)

internal/core/commands/update_dd_image.go
```diff
@@ -0,0 +1,76 @@
...
+				Err: err,
+			}
+		}
+	}
+
+	if err != nil {
+		return nil, &exceptions.NotFoundError{
+			Err: fmt.Errorf("could not find device definition id: %s", command.DeviceDefinitionID),
+		}
+	}
```
JamesReate, line 60: reorganize logic eg. 
if err != nil {
    if errors.Is(err, sql.ErrNoRows) { 
       return nil, &exceptions.NotFoundError{
 			Err: fmt.Errorf("could not find device definition id: %s", command.DeviceDefinitionID),
 		}
    }
    return nil, &exceptions.InternalError{
 				Err: err,
 			}
}

JamesReate (approved):

PR #16, opened 2022-09-17
add device definition dynamic filter endpoint

Asfiroth:
# Proposed Changes
- Add Endpoint for admin dynamic filtering


internal/api/handlers/device_definition_handler.go
```diff
@@ -75,3 +76,37 @@ func GetDeviceDefinitionByMMY(m mediator.Mediator) fiber.Handler {
 		return c.Status(fiber.StatusOK).JSON(result)
 	}
 }
+
+func GetDeviceDefinitionByDynamicFilter(m mediator.Mediator) fiber.Handler {
```
JamesReate, line 80: do we need a fiber handler since this would be done over grpc? eg. no need for a rest endpoint

internal/api/routes.go
```diff
@@ -11,6 +11,7 @@ func RegisterDeviceDefinitionsRoutes(app fiber.Router, m mediator.Mediator) {
 	app.Get("/device-definitions/:id", handlers.GetDeviceDefinitionByID(m))
 	app.Get("/device-definitions/:id/integrations", handlers.GetDeviceIntegrationsByID(m))
 	app.Get("/device-definitions", handlers.GetDeviceDefinitionByMMY(m))
+	app.Get("/device-definitions/filter", handlers.GetDeviceDefinitionByDynamicFilter(m))
```
JamesReate, line 14: re: see above comment - not sure if we need this?

rhvivancoeffio (approved):

internal/api/handlers/device_definition_handler.go
Asfiroth: I think we don't 🤔 ...

PR #18, opened 2022-09-19
EXP-57 Fetch features data from es and setup database model for persisting features in db

0xdev22:
# Proposed Changes
The feature is a script that will help get device features from elasticsearch and persist said data in the database. This PR contains 
- The initial script setup
- The es aggregation query to fetch device features from
- The model for storing features in db
- Adding the database column for features to devices_integration


pkg/elastic/elastic_search_api_service.go
```diff
@@ -29,7 +29,7 @@ type SearchService interface {
 	GetMetaEngineName() string
 }
 
-type elasticSearchService struct {
+type ElasticSearchService struct {
```
elffjs, line 32: Why'd we make this public?

.gitignore
```diff
@@ -18,4 +18,5 @@
 settings.yaml
 
 .idea/
-data/
\ No newline at end of file
+data/
+.vscode
```
elffjs, line 22: Probably a good change!

pkg/elastic/elastic_search_api_service.go
```diff
@@ -41,7 +41,20 @@ func NewElasticSearchService(settings *config.Settings, logger zerolog.Logger) (
 	var netClient = &http.Client{
 		Timeout: time.Second * 10,
 	}
-	return &elasticSearchService{
+	return &ElasticSearchService{
```
elffjs, line 44: Hmm, I think we must be careful here. "App search" is the separate thing that powers the model autocomplete in the app. You want the "analytics host", I think. Might be best as a separate client? They don't have much to do with each other.

elffjs:
It makes the most sense to me to do the PR that adds the table with all the signals first, and then to drive these queries from that.

pkg/elastic/elastic_search_api_service.go
JamesReate: this client is for the enterprise app search, so you won't want to use it.
When you create the new client for the regular ES, don't put it in the pkg folder though, put it in infrastructure

pkg/elastic/elastic_search_api_service.go
JamesReate: Correct, you'll need to add different settings too. 
[Example here](https://github.com/DIMO-Network/device-data-api/blob/e3db1632083d7df73c0fb2cf319c1397876e1b63/internal/config/settings.go#L11-L14), and you can trace back the settings for the client - might be a good one to put into the [shared](https://github.com/DIMO-Network/shared) repo.

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -2,16 +2,45 @@ package main
...
 	"github.com/DIMO-Network/device-definitions-api/internal/config"
 	"github.com/DIMO-Network/device-definitions-api/internal/infrastructure/db/models"
 	"github.com/DIMO-Network/device-definitions-api/pkg/elastic"
 	"github.com/DIMO-Network/shared/db"
 	"github.com/rs/zerolog"
+	"github.com/volatiletech/sqlboiler/v4/boil"
 )
 
+func prepareFeatureData(f map[string]map[string]int) []models.DeviceIntegrationFeatures {
+	displayNameMapping := map[string]string{
```
elffjs, line 16: I think we want this to be a database table.

internal/infrastructure/db/migrations/20220922010142_integration_features.sql
```diff
@@ -0,0 +1,34 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+CREATE TABLE IF NOT EXISTS device_definitions_api.integration_features
+(
+    elastic_property    varchar(50) PRIMARY KEY not null,
+    feature_key         varchar(50) not null,
```
elffjs, line 7: This seems like the primary key to me. What do you think?

internal/infrastructure/db/migrations/20220922010142_integration_features.sql
```diff
@@ -0,0 +1,34 @@
...
+    ('tires', 'Tires', 'tires'),
+    ('speed', 'Speed', 'speed');
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+SELECT 'down SQL query';
+
+DROP INDEX feature_key_idx;
+DROP INDEX css_icon_idx;
```
elffjs, line 31: `DROP TABLE` should take care of these.

.gitignore
```diff
@@ -19,6 +19,7 @@ settings.yaml
 
 .idea/
 data/
+.vscode
```
elffjs, line 22: Good change.

internal/infrastructure/elastic_search/elastic_search_service.go
```diff
@@ -0,0 +1,113 @@
...
+	log            zerolog.Logger
+}
+
+func NewElasticSearch(settings *config.Settings, logger zerolog.Logger) (*ElasticSearch, error) {
+	var netClient = &http.Client{
+		Timeout: time.Second * 10,
+	}
+	return &ElasticSearch{
+		BaseURL:        settings.ElasticSearchAppSearchHost,
+		Token:          settings.ElasticSearchAppSearchToken,
```
elffjs, line 31: We'll want to create settings with more appropriate names (we're not using "app search").

internal/infrastructure/elastic_search/elastic_search_service.go
```diff
@@ -0,0 +1,113 @@
...
+		MetaEngineName: "dd-" + settings.Environment,
+		httpClient:     netClient,
+		log:            logger,
+	}, nil
+}
+
+// buildAndExecRequest makes request with token and headers, marshalling passed in obj or if string just passing along in body,
+// and unmarshalling response body to objOut (must be passed in as ptr eg &varName)
+func (d *ElasticSearch) buildAndExecRequest(method, url string, obj interface{}, objOut interface{}) (*http.Response, error) {
```
elffjs, line 40: I wonder if we could add something like this to the shared library. There's an HTTP client in there with retry and backoff but it doesn't do marshaling.

internal/infrastructure/elastic_search/query_device_features.go
```diff
@@ -0,0 +1,141 @@
+package elasticsearch
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
+type DeviceFeatures struct {
```
elffjs, line 9: Are we using this?

internal/infrastructure/elastic_search/query_device_features.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	Tires struct {
+		DocumentCount int64 `json:"doc_count"`
+	}
+}
+
+type DeviceFeaturesResp struct {
+	Aggregations struct {
+		Features struct {
+			SumOtherDocCount int64 `json:"sum_other_doc_count"`
```
elffjs, line 36: Are we doing anything with these numbers? Hopefully we've set `size` to be so high that this is always `0`.

internal/infrastructure/elastic_search/query_device_features.go
```diff
@@ -0,0 +1,141 @@
...
+	Speed struct {
+		DocumentCount int64 `json:"doc_count"`
+	}
+	Tires struct {
+		DocumentCount int64 `json:"doc_count"`
+	}
+}
+
+type DeviceFeaturesResp struct {
+	Aggregations struct {
```
elffjs, line 34: I had forgotten that `encoding/json` will match `"aggregations"` to `Aggregations`. That is nice.

internal/infrastructure/elastic_search/query_device_features.go
```diff
@@ -0,0 +1,141 @@
...
+				},
+				"aggs": {
+			  "features":{
+				"filters": {
+				  "filters": {
+					"odometer":{
+					  "exists": {
+						"field": "data.odometer"
+					  }
+					},
```
elffjs, line 91: Here's the interesting part, having combined what might have been two PRs: I think you want to construct this section using the `device_features` table. So now it's (using pseudocode)
```
for row in device_integrations:
    filters[row.feature_key] = {"exists": {"field": "data." + row.elastic_field}}
```

internal/infrastructure/elastic_search/query_device_features.go
```diff
@@ -0,0 +1,141 @@
...
+				}
+			}
+		}
+	}
+}
+
+func (d *ElasticSearch) GetDeviceFeatures(envName string) (DeviceFeaturesResp, error) {
+	url := fmt.Sprintf("%s/device-status-%s*/_search", d.BaseURL, envName)
+	body := `
+	{
```
elffjs, line 58: If we are keeping this blob, might we run it through a formatter?

internal/infrastructure/elastic_search/query_device_features.go
```diff
@@ -0,0 +1,141 @@
...
+		Features struct {
+			SumOtherDocCount int64 `json:"sum_other_doc_count"`
+			Buckets          []struct {
+				Key               string `json:"key"`
+				DocumentCount     int64  `json:"doc_count"`
+				DeviceDefinitions struct {
+					SumOtherDocCount int64 `json:"sum_other_doc_count"`
+					Buckets          []struct {
+						Key           string `json:"key"`
+						DocumentCount int64  `json:"doc_count"`
```
elffjs, line 44:
```suggestion
						DocCount int64  `json:"doc_count"`
```
Just to match `SumOtherDocCount`. Not a big deal.

internal/infrastructure/db/migrations/20220919141609_device_integration_features.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
```
elffjs, line 3: By the way, these `SELECT 'up SQL query';` lines don't do anything. I think a lot of people leave them in, though.

internal/infrastructure/elastic_search/query_device_features.go
```diff
@@ -0,0 +1,141 @@
...
+						Features      struct {
+							Buckets map[string]map[string]int
```
elffjs, line 46: It might be worth it to have this be
```suggestion
							Buckets map[string]struct{
								DocCount int `json:"doc_count"`
							}
```

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,88 @@
...
+
+	es, _ := elastic.NewElasticSearch(s, logger)
+
+	resp, err := es.GetDeviceFeatures(s.Environment)
+	if err != nil {
+		return err
+	}
+
+	for _, f := range resp.Aggregations.Features.Buckets {
+		intID := strings.Split(f.Key, "/")[2]
```
elffjs, line 45: In the past I've done `intID := strings.TrimPrefix(f.Key, "dimo/integration/")`.

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,88 @@
...
+
+		for _, d := range f.DeviceDefinitions.Buckets {
+			ddID := d.Key
+
+			devices, err := models.DeviceIntegrations(
+				models.DeviceIntegrationWhere.DeviceDefinitionID.EQ(ddID),
+				models.DeviceIntegrationWhere.IntegrationID.EQ(intID),
+			).All(ctx, pdb.DBS().Reader)
```
elffjs, line 53: So, the primary key on this table is `(device definition id, integration id, region name)`. With this I'm not sure what region you'll get—could we update all of them?

internal/infrastructure/db/models/integration_features.go
```diff
@@ -0,0 +1,13 @@
+package models
+
+type IntegrationFeatures struct {
```
elffjs, line 3: Are we using this?

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,88 @@
...
+				// handle when we have region in elasticsearch
+			} */
+
+			if len(devices) < 1 {
+				// handle not found
+				logger.Error().Msgf("error could not find device with integration id %s and deviceDefinitionId %s", intID, ddID)
+				continue
+			}
+
+			feature := prepareFeatureData(d.Features.Buckets)
```
elffjs, line 70: In this JSON field it seems like we're putting
```
[{"elastic_key": "soc", "supportLevel": 2}, …]
```
If you want to use the Elastic key (I might have called it field, I think the Elastic people usually talk about fields) I think that's fine. In that case we can probably dump the `feature_key` column, then.

It would be nice to be consistent with casing, so writing `elasticKey` here.

What does support level 2 mean?

internal/infrastructure/db/models/integration_features.go
```diff
@@ -0,0 +1,13 @@
...
+	FeatureKey      string `json:"feature_key,omitempty"`
+	DisplayName     string `json:"display_name,omitempty"`
+	IconCSS         string `json:"css_icon,omitempty"`
+	ElasticProperty string `json:"elastic_property,omitempty"`
+}
+
+type DeviceIntegrationFeatures struct {
```
elffjs, line 10: I think we probably only want SQLBoiler models in this folder—typically we run `sqlboiler psql --no-tests --wipe`, which deletes the whole `models` folder.

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,88 @@
...
+		for _, d := range f.DeviceDefinitions.Buckets {
+			ddID := d.Key
+
+			devices, err := models.DeviceIntegrations(
+				models.DeviceIntegrationWhere.DeviceDefinitionID.EQ(ddID),
+				models.DeviceIntegrationWhere.IntegrationID.EQ(intID),
+			).All(ctx, pdb.DBS().Reader)
+			// check if device exists
+			if err != nil {
+				logger.Error().Msgf("error occurred fetching device with integration id %s and deviceDefinitionId %s. error: %s", intID, ddID, err.Error())
```
elffjs, line 56: Minor nitpick: we're not fetching a device, exactly. Compatibility record?

elffjs: In general I would write a log line like this as
```suggestion
				logger.Err(err).Str("integrationId", indID).Str("deviceDefinitionId", ddID).("error occurred fetching device")
```
It's easier to search and filter in Elastic and is actually faster! None of this is a huge concern here, but in general it's helpful.

internal/infrastructure/elastic_search/query_device_features.go
0xdev22: Yes that makes sense..

internal/infrastructure/elastic_search/query_device_features.go
0xdev22: We have a high enough number,  but I'm including it as a precaution. I intend to check it and log an error if the number is ever above 0

internal/infrastructure/elastic_search/query_device_features.go
0xdev22: The structure of the response is like this 

``` json
"buckets": {
  "batteryVoltage": {
	  "doc_count": 0
  },
  "fuelPercentRemaining": {
	  "doc_count": 0
  },
  "odometer": {
	  "doc_count": 0
  },
...
```

internal/infrastructure/elastic_search/query_device_features.go
elffjs: That would be really nice! We should probably do that elsewhere.

If it does get too large we can kinda keep limping along with [the composite aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-composite-aggregation.html), which does paginate I think.

There's also [collapsing](https://www.elastic.co/guide/en/elasticsearch/reference/current/collapse-search-results.html), which I haven't looked at yet.

internal/infrastructure/elastic_search/query_device_features.go
elffjs: Are we disagreeing? Maybe I'm too sleepy.

cmd/device-definitions-api/populate_device_features.go
0xdev22: As discussed on this topic, it will be handled once we handle region data for es

internal/infrastructure/db/models/integration_features.go
0xdev22: Yes, its being used heavily `populate_device_features` https://github.com/DIMO-Network/device-definitions-api/pull/18/commits/7e52deb44f552d8598ac06fce156014064373a0a#diff-e1372ebd61844d8cc4bba372a80c46eeb7de4f26df27b6c040bce1a1f1f643beR65

cmd/device-definitions-api/populate_device_features.go
0xdev22: Noted and corrected about the key consistency and also the primary key. The support levels for this task is 

0 - Not supported
1 - Maybe Supported
2 - Supported

internal/infrastructure/db/models/integration_features.go
elffjs: Forgive me, I don’t see the need for a raw query there.

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/volatiletech/sqlboiler/v4/boil"
+)
+
+type jsonObj map[string]interface{}
+
+func prepareFeatureData(f map[string]map[string]int) []elasticModels.DeviceIntegrationFeatures {
+	ft := []elasticModels.DeviceIntegrationFeatures{}
+
+	for k, v := range f {
+		supportLevel := 0
```
elffjs, line 23:
```suggestion
		var supportLevel int8
```
Then you can avoid all this casting, I think. But I wonder if this is worth it.

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/DIMO-Network/device-definitions-api/internal/config"
+	"github.com/DIMO-Network/device-definitions-api/internal/infrastructure/db/models"
+	elastic "github.com/DIMO-Network/device-definitions-api/internal/infrastructure/elasticsearch"
+	elasticModels "github.com/DIMO-Network/device-definitions-api/internal/infrastructure/elasticsearch/models"
+	"github.com/DIMO-Network/shared/db"
+	"github.com/rs/zerolog"
+	"github.com/volatiletech/sqlboiler/v4/boil"
+)
+
+type jsonObj map[string]interface{}
```
JamesReate, line 17:
```suggestion
type jsonObj map[string]any
```

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,123 @@
...
+
+func prepareFeatureData(f map[string]map[string]int) []elasticModels.DeviceIntegrationFeatures {
+	ft := []elasticModels.DeviceIntegrationFeatures{}
+
+	for k, v := range f {
+		var supportLevel int8
+
+		if v["doc_count"] > 0 {
+			supportLevel = 2
```
JamesReate, line 26: It is hard to know what `2` means here. I think this could be a good use case for an "enum" type (which go doesn't support natively) but you can do something like [this example](https://github.com/DIMO-Network/devices-api/blob/fac677e864105e97cae3315a45ba2e7a64a0a0a0/internal/services/ingest_registrar.go#L98-L107)

internal/infrastructure/db/migrations/20220922010142_integration_features.sql
```diff
@@ -0,0 +1,34 @@
...
+(
+    feature_key         varchar(50) PRIMARY KEY not null,
+    elastic_property    varchar(50) not null,
+    display_name        varchar(50) not null,
+    css_icon            varchar(100),
+    created_at          timestamptz not null default current_timestamp,
+    updated_at          timestamptz not null default current_timestamp
+);
+CREATE UNIQUE INDEX elastic_property_idx ON device_definitions_api.integration_features (elastic_property);
+CREATE UNIQUE INDEX css_icon_idx ON device_definitions_api.integration_features (css_icon);
```
JamesReate, line 14: I'm not sure I would enforce unique on the css icon, i could see them re-using an icon idk.

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,123 @@
...
+
+	esFilters, err := json.Marshal(filters)
+	if err != nil {
+		return "", err
+	}
+
+	return string(esFilters), nil
+}
+
+func populateDeviceFeaturesFromEs(ctx context.Context, logger zerolog.Logger, s *config.Settings) error {
```
JamesReate, line 64: Consider moving this method to the top as it does the meat of the operations, and the other methods act more like helpers / refactor out logic .

cmd/device-definitions-api/populate_device_features.go
```diff
@@ -0,0 +1,123 @@
...
+				continue
+			}
+
+			/* if len(devices) > 1 { // we have for multiple continents
+				// handle when we have region in elasticsearch
+			} */
+
+			if len(devices) < 1 {
+				// handle not found
+				logger.Error().Msgf("error could not find device with integration id %s and deviceDefinitionId %s", intID, ddID)
```
JamesReate, line 102: let's make these log messages also use the `.Str(..)` like the one on line 92 for well known properties (integrationId, deviceDefinitionId etc).

JamesReate:
Let's also update readme with a new section on how to run this locally. The tricky thing will be test data - but you can look at example in dimo-explorer or ask @Allyson-English how she managed with that. I think it may be a problem we have solve before.

0xdev22:
> Let's also update readme with a new section on how to run this locally. The tricky thing will be test data - but you can look at example in dimo-explorer or ask @Allyson-English how she managed with that. I think it may be a problem we have solve before.

I actually have my local test environment setup and I was able to test this before pushing. We might need to write something more comprehensive in a central location on how to setup local testing with es and how to generate test data.

JamesReate:
> > Let's also update readme with a new section on how to run this locally. The tricky thing will be test data - but you can look at example in dimo-explorer or ask @Allyson-English how she managed with that. I think it may be a problem we have solve before.
> 
> I actually have my local test environment setup and I was able to test this before pushing. We might need to write something more comprehensive in a central location on how to setup local testing with es and how to generate test data.

That's great. 
For central docs, we can use the [internal docs](https://github.com/DIMO-Network/internal-docs) repo - there's already some elastic stuff in there or start a separate one specific to this. Anything repo specific is fine in the Readme.

JamesReate (approved):

internal/infrastructure/elastic_search/query_device_features.go
0xdev22: Ok.. Since the number is high, i'm thinking of leaving this out in this PR..

internal/infrastructure/elasticsearch/elastic_search_service.go
```diff
@@ -0,0 +1,111 @@
...
+type ElasticSearch struct {
+	BaseURL    string
+	Token      string
+	httpClient *http.Client
+	log        zerolog.Logger
+}
+
+func NewElasticSearch(settings *config.Settings, logger zerolog.Logger) (*ElasticSearch, error) {
+	var netClient = &http.Client{
+		Timeout: time.Second * 10,
```
elffjs, line 26: Worried slightly about this timeout.

PR #23, opened 2022-09-23
Device definition api validate arrays

rhvivancoeffio:
# Proposed Changes

### DIFF
--- a/internal/core/queries/get_dd_by_ids.go
+++ b/internal/core/queries/get_dd_by_ids.go
@@ -7,8 +7,11 @@ import (
 
 	"github.com/DIMO-Network/device-definitions-api/internal/core/models"
 	"github.com/DIMO-Network/device-definitions-api/internal/core/services"
+	"github.com/DIMO-Network/device-definitions-api/internal/infrastructure/exceptions"
 	"github.com/DIMO-Network/device-definitions-api/pkg/grpc"
 	"github.com/TheFellow/go-mediator/mediator"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
 )
 
 type GetDeviceDefinitionByIdsQuery struct {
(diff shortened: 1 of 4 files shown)


pkg/dbtest/test_helper.go
```diff
@@ -191,3 +191,11 @@ func SetupCreateDeviceIntegration(t *testing.T, dd *models.DeviceDefinition, int
 	assert.NoError(t, err)
 	return di
 }
+
+func Logger() *zerolog.Logger {
+	l := zerolog.New(os.Stdout).With().
+		Timestamp().
+		Str("app", "devices-api").
```
JamesReate, line 198:
```suggestion
		Str("app", "device-definitions-api").
```

internal/core/queries/get_dd_by_ids.go
```diff
@@ -19,25 +22,33 @@ func (*GetDeviceDefinitionByIdsQuery) Key() string { return "GetDeviceDefinition
...
+	}
+
 	response := &grpc.GetDeviceDefinitionResponse{}
 
 	for _, v := range qry.DeviceDefinitionID {
 		dd, _ := ch.DDCache.GetDeviceDefinitionByID(ctx, v)
 
 		if dd == nil {
-			fmt.Printf("Not found")
+			ch.log.Warn().Msg(fmt.Sprintf("Device Definition Id %s not found", v))
```
JamesReate, line 51: I liked the log message starting with "Not found" because made it easy to see in logs. Consider changing to:
`ch.log.Warn().Msg(fmt.Sprintf("Not found - deviceDefinitionId %s", v))`

elffjs: I like putting the variable stuff in the properties too.

PR #28, opened 2022-09-27
Return device make token id

rhvivancoeffio:
# Proposed Changes
Return device make token id

### Impacted Routes

GRPC Endpoints
GetDeviceDefinitionByID
GetDeviceDefinitionByMMY


pkg/grpc/device_definition.proto
```diff
@@ -67,6 +67,7 @@ message GetDeviceDefinitionItemResponse {
 		string name = 2;
 		string logo_url = 3;
 		string oem_platform_name = 4;
+		uint64 token_id = 5;
```
elffjs, line 70: Are we to interpret this being 0 as it being absent? I feel like we're losing something.

PR #30, opened 2022-09-28
PLA-176 Enable prometheus metrics

rhvivancoeffio:
# Proposed Changes
Enable prometheus metrics

### DIFF
--- a/internal/api/common/behaviors.go
+++ b/internal/api/common/behaviors.go
@@ -3,6 +3,7 @@ package common
 import (
 	"context"
 	"fmt"
+
 	"github.com/DIMO-Network/device-definitions-api/internal/config"
 	"github.com/DIMO-Network/device-definitions-api/internal/infrastructure/exceptions"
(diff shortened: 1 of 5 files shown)


internal/api/common/behaviors.go
```diff
@@ -42,40 +42,52 @@ func (p ValidationBehavior) Process(ctx context.Context, msg mediator.Message, n
...
+			panic(err)
+		}
+
+		if _, ok := err.(*exceptions.NotFoundError); ok {
+			metrics.NotFoundRequestError.With(prometheus.Labels{"method": msg.Key()}).Inc()
+			panic(err)
+		}
+
+		metrics.InternalError.With(prometheus.Labels{"method": msg.Key()}).Inc()
 		panic(err)
```
JamesReate, line 86: we can remove above `panic(err)` in the `if` statements since we always panic regardless here

JamesReate (approved):
just address couple comments and good to merge

PR #37, opened 2022-09-30
EXP-113 Implement grpc endpoint to fetch device compatibility data by makeID

0xdev22:
# Proposed Changes
This PR is a follow up to the #18 . We introduce a new grpc endpoint that takes the features/compatibility data that was saved to es and returns the info for all models under that makeID.

> Please note that the migrations for integration_features table and device_integration tables needs to be run first so the necessary schema are created

### Impacted Routes
NA
### Caveats
NA


internal/api/grpc_service.go
```diff
@@ -322,3 +324,31 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *p_grpc.Emp
 
 	return result, nil
 }
+
+func (s *GrpcService) GetDeviceCompaitibility(ctx context.Context, in *p_grpc.GetDeviceCompatibilityListRequest) (*p_grpc.GetDeviceCompatibilityListResponse, error) {
+	qryResult, _ := s.Mediator.Send(ctx, &queries.GetDeviceCompatibilityQuery{
+		MakeID: in.MakeId,
+	})
+	// log.Println(qryResult, "qryResultqryResult")
```
JamesReate, line 332: nit: remove debug comment

pkg/grpc/device_definition.proto
```diff
@@ -260,6 +262,21 @@ message CreateDeviceStyleResponse {
...
+
+message DeviceCompaitibilityList {
+	string model = 1;
+	int32 year = 2;
+	google.protobuf.Struct features = 3;
+}
+
+message GetDeviceCompatibilityListResponse {
+	string make = 1;
+	repeated DeviceCompaitibilityList deviceCompatilities = 2;
```
JamesReate, line 277: I know we had a mix of camelCase and lower_snake_case, but if you see the latest PR we're trying to keep everything standard to `lower_snake_case`
```suggestion
	repeated DeviceCompaitibilityList device_compatilities = 2;
```

internal/api/grpc_service.go
```diff
@@ -322,3 +323,26 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *p_grpc.Emp
 
 	return result, nil
 }
+
+func (s *GrpcService) GetDeviceCompaitibility(ctx context.Context, in *p_grpc.GetDeviceCompatibilityListRequest) (*p_grpc.GetDeviceCompatibilityListResponse, error) {
```
JamesReate, line 327:
```suggestion
func (s *GrpcService) GetDeviceCompatibility(ctx context.Context, in *p_grpc.GetDeviceCompatibilityListRequest) (*p_grpc.GetDeviceCompatibilityListResponse, error) {
```

pkg/grpc/device_definition.proto
```diff
@@ -260,6 +262,21 @@ message CreateDeviceStyleResponse {
 	string ID = 1;
 }
 
+message GetDeviceCompatibilityListRequest {
+	string makeId = 1;
+}
+
+message DeviceCompaitibilityList {
```
JamesReate, line 269:
```suggestion
message DeviceCompatibilityList {
```

pkg/grpc/device_definition.proto
```diff
@@ -260,6 +262,21 @@ message CreateDeviceStyleResponse {
...
+	string model = 1;
+	int32 year = 2;
+	google.protobuf.Struct features = 3;
+}
+
+message GetDeviceCompatibilityListResponse {
+	string make = 1;
+	repeated DeviceCompaitibilityList device_compatilities = 2;
```
JamesReate, line 277:
```suggestion
	repeated DeviceCompaitibilityList device_compatibilities = 2;
```

internal/core/queries/device_compatibility.go
```diff
@@ -0,0 +1,85 @@
...
+			return GetDeviceCompatibilityQueryResult{}, nil
+		}
+		for _, i := range dd {
+			f := i.(map[string]interface{})
+			fk := f["feature_key"]
+			sl := f["support_level"]
+			feats[integFeats[fk.(string)]] = sl.(float64) > 0
+		}
+
+		cr.Features = feats
```
JamesReate, line 79: I wonder if to try to maintain type safety, for each feature you could return a struct something like
```go
type Feature struct {
  Key string // eg. odometer
  SupportLevel int // eg. 0,1,2
}
```

0xdev22: Was initially thinking about keeping things flexible, but we can also go the route of a more structured type

elffjs: Structuring seems good

pkg/grpc/device_definition.proto
```diff
@@ -241,6 +241,31 @@ message GetDeviceStyleByIDRequest {
...
+	string integration_id = 2;
+	string region = 3;
+}
+
+message Feature {
+	string key = 1;
+	int32 support_level = 2;
+}
+
+message DeviceModelYears {
```
elffjs, line 255:
```suggestion
message DeviceModelYear {
```

elffjs: Well, apart from being plural or not maybe this ought to be more specific? Maybe "compatibility" ought to be in the name.

internal/infrastructure/db/repositories/device_definition_repo.go
```diff
@@ -180,3 +181,24 @@ func (r *deviceDefinitionRepository) GetOrCreate(ctx context.Context, source str
...
+		qm.Where("device_definitions.device_make_id = ?", makeID),
+		qm.Where("device_definitions.year >= ?", 2008),
+		qm.Where("device_integrations.features IS NOT NULL"),
+		qm.Where("device_integrations.integration_id = ?", integrationID),
+		qm.Where("device_integrations.region = ?", region),
+		qm.Load(models.DeviceDefinitionRels.DeviceIntegrations),
+	).All(ctx, r.DBS().Reader)
+
+	if err != nil {
+		if !errors.Is(err, sql.ErrNoRows) {
```
elffjs, line 197: I believe one can't get `ErrNoRows` out of `All`.

internal/infrastructure/db/repositories/device_definition_repo.go
```diff
@@ -180,3 +181,24 @@ func (r *deviceDefinitionRepository) GetOrCreate(ctx context.Context, source str
 	}
 	return dd, nil
 }
+
+func (r *deviceDefinitionRepository) FetchDeviceCompatibility(ctx context.Context, makeID, integrationID, region string) (models.DeviceDefinitionSlice, error) {
+	res, err := models.DeviceDefinitions(
+		qm.InnerJoin("device_integrations ON device_definitions.id = device_integrations.device_definition_id"),
```
elffjs, line 187: There are some examples of doing this with the constants [here](https://github.com/volatiletech/sqlboiler#query-mod-system). Maybe it's not worth it?

internal/infrastructure/db/repositories/device_definition_repo.go
```diff
@@ -180,3 +181,24 @@ func (r *deviceDefinitionRepository) GetOrCreate(ctx context.Context, source str
...
+		qm.Where("device_definitions.device_make_id = ?", makeID),
+		qm.Where("device_definitions.year >= ?", 2008),
+		qm.Where("device_integrations.features IS NOT NULL"),
+		qm.Where("device_integrations.integration_id = ?", integrationID),
+		qm.Where("device_integrations.region = ?", region),
```
elffjs, line 192: These we might be able to do with the `…Where`. Up to you!

internal/api/grpc_service.go
```diff
@@ -443,6 +442,62 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+	result := &p_grpc.GetDeviceCompatibilityListResponse{}
+
+	integFeats := deviceCompatibilities.IntegrationFeatures
+	dcMap := make(map[string][]*p_grpc.DeviceModelYears)
+	for _, v := range deviceCompatibilities.DeviceDefinitions {
+		// struct has noa values yet
+		if _, ok := dcMap[v.Model]; !ok {
+			dcMap[v.Model] = []*p_grpc.DeviceModelYears{}
+		}
+		if len(v.R.DeviceIntegrations) == 0 {
```
elffjs, line 463: Shouldn't happen because of the inner join, yeah? Maybe a warning?

internal/api/grpc_service.go
```diff
@@ -443,6 +442,62 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+			continue
+		}
+
+		di := v.R.DeviceIntegrations[0]
+
+		if di.Features.IsZero() {
+			continue
+		}
```
elffjs, line 471: We shouldn't hit any of these because of the query, yeah? Maybe worth a warning?

internal/api/grpc_service.go
```diff
@@ -443,6 +442,62 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+		res := &p_grpc.DeviceModelYears{Year: int32(v.Year)}
+
+		feats := []*p_grpc.Feature{}
+		var dd []interface{}
```
elffjs, line 475:
```suggestion
		var features []struct{
			FeatureKey   string
			SupportLevel int32
		}
```

internal/api/grpc_service.go
```diff
@@ -443,6 +442,62 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+
+	result := &p_grpc.GetDeviceCompatibilityListResponse{}
+
+	integFeats := deviceCompatibilities.IntegrationFeatures
+	dcMap := make(map[string][]*p_grpc.DeviceModelYears)
+	for _, v := range deviceCompatibilities.DeviceDefinitions {
+		// struct has noa values yet
+		if _, ok := dcMap[v.Model]; !ok {
+			dcMap[v.Model] = []*p_grpc.DeviceModelYears{}
+		}
```
elffjs, line 462: If you remove these lines, does anything go wrong?

pkg/grpc/device_definition.proto
```diff
@@ -283,6 +308,7 @@ service DeviceDefinitionService {
 	rpc GetDeviceMakeByName(GetDeviceMakeByNameRequest) returns (DeviceMake);
 	rpc GetDeviceMakes(google.protobuf.Empty) returns (GetDeviceMakeResponse);
 	rpc GetFilteredDeviceDefinition(FilterDeviceDefinitionRequest) returns (GetFilteredDeviceDefinitionsResponse);
+	rpc GetDeviceCompatibility(GetDeviceCompatibilityListRequest) returns (GetDeviceCompatibilityListResponse);
```
elffjs, line 311: I wonder if the name conveys well enough what this does. When I see `GetDeviceCompatibility` I'm thinking singular.

internal/api/grpc_service.go
```diff
@@ -443,6 +442,62 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+		err := di.Features.Unmarshal(&dd)
+		if err != nil {
+			continue
+		}
+
+		for _, i := range dd {
+			f := i.(map[string]interface{})
+			ft := &p_grpc.Feature{}
+			ft.Key = integFeats[f["feature_key"].(string)]
+			ft.SupportLevel = int32(f["support_level"].(float64))
```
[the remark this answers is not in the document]
elffjs: Actually, can we just re-use the struct we used to put these in the DB? That’s probably the safest thing.

I’m not sure this will work, as written with the snake case.

internal/api/grpc_service.go
```diff
@@ -7,7 +7,6 @@ import (
 
 	"github.com/DIMO-Network/device-definitions-api/internal/core/commands"
 	"github.com/DIMO-Network/device-definitions-api/internal/core/models"
-	coremodels "github.com/DIMO-Network/device-definitions-api/internal/core/models"
```
elffjs, line 10: This seems like a good change. I wonder if the renaming was meant to avoid confusion with the SQLBoiler models.

internal/api/grpc_service.go
```diff
@@ -443,6 +442,62 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+		if di.Features.IsZero() {
+			continue
+		}
+		res := &p_grpc.DeviceModelYears{Year: int32(v.Year)}
+
+		feats := []*p_grpc.Feature{}
+		var dd []interface{}
+		err := di.Features.Unmarshal(&dd)
+		if err != nil {
+			continue
```
elffjs, line 478: Similarly, it might be worth logging.

internal/api/grpc_service.go
0xdev22: Actually will still work, just defensive programming I guess

internal/api/grpc_service.go
0xdev22: The problem is models is still used predominantly here, coremodels was used by just two functions I think. I can make a new branch to re-introduce coremodels because changing all of them in this PR will introduce a lot of irrelevant changes

elffjs: I think it’s perfectly fine the way you have it.

internal/api/grpc_service.go
```diff
@@ -443,6 +445,75 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+		Region:        in.Region,
+	})
+
+	deviceCompatibilities := qryResult.(queries.GetDeviceCompatibilityQueryResult)
+
+	result := &p_grpc.GetDeviceCompatibilityListResponse{}
+
+	integFeats := deviceCompatibilities.IntegrationFeatures
+	dcMap := make(map[string][]*p_grpc.DeviceCompatibilities)
+	for _, v := range deviceCompatibilities.DeviceDefinitions {
```
JamesReate, line 461: nit: thoughts on naming `v` `dd`

internal/api/grpc_service.go
```diff
@@ -443,6 +445,75 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+				Str("Model", v.Model).
+				Str("DeviceDefinition", v.ID).
+				Msg("Error de-serializing features information")
+			continue
+		}
+
+		for _, f := range features {
+			fkey := f.FeatureKey
+			if fkey == "tires" {
+				fkey = "tires.frontLeft"
```
JamesReate, line 496: I know we know why this is the case, but good place to have a comment for future ref

pkg/grpc/device_definition.proto
```diff
@@ -241,6 +241,31 @@ message GetDeviceStyleByIDRequest {
 	string id = 1;
 }
 
+message GetDeviceCompatibilityListRequest {
+	string make_id = 1;
+	string integration_id = 2;
+	string region = 3;
+}
+
+message Feature {
```
JamesReate, line 250: I wonder if we should make `Feature` an embedded type of DeviceCompatibilities since it only lives in the context of that. The drawback is that it makes the generated type name longer. Not sure what you guys think.

internal/api/grpc_service.go
elffjs: I'm not sure why this internal endpoint beneath this call is exposing the Elastic column name. Why not use the feature key everywhere where you're not directly hitting Elastic?

internal/api/grpc_service.go
```diff
@@ -443,6 +445,75 @@ func (s *GrpcService) GetDeviceDefinitionAll(ctx context.Context, in *emptypb.Em
...
+			}
+			ft := &p_grpc.Feature{
+				Key:          integFeats[fkey],
```
elffjs, line 499: Why call this "key" if it's really the display name?

pkg/grpc/device_definition.proto
0xdev22: I think this is a good idea, it will definitely make things easier to follow

internal/api/grpc_service.go
0xdev22: It was a convention we agreed on initially. We can make a change in another PR

pkg/grpc/device_definition.proto
elffjs: I think embedding some of these would assuage my concerns about the names being not descriptive enough. Too late now, I suppose.

PR #38, opened 2022-10-01
Rename struct and protos

rhvivancoeffio:
# Proposed Changes
Rename structs and proto messages


JamesReate (approved):
I think it's good, standardizing property names to snake_lowercase, so they're all the same and so they match what was done in the `shared` repo for other grpc services.

PR #43, opened 2022-10-06
EXP-122 sync nhtsa recalls

zavaboy:
# Proposed Changes
Added `nhtsa-sync-recalls` command
Inserts into `device_nhtsa_recalls` table
Added device definition matching
Added tests to ensure recall data is properly inserted, updated to include matched DD and matching info for each type of matching

### Impacted Routes
New `nhtsa-sync-recalls` command

### Caveats
DD matching will be imperfect


internal/config/settings.go
```diff
@@ -24,4 +24,5 @@ type Settings struct {
 	DrivlyVINAPIURL                string         `yaml:"DRIVLY_VIN_API_URL"`
 	DrivlyOfferAPIURL              string         `yaml:"DRIVLY_OFFER_API_URL"`
 	Redis                          redis.Settings `yaml:"REDIS"`
+	NHTSARecallsFileURL            string         `yaml:"NHTSA_RECALLS_FILE_URL"`
```
JamesReate, line 27: add a setting for the kubernetes to pull for this in `/charts/device-definitions-api/values.yaml` under the `env:` section. Since there isn't a different value for production you only need to set it there.

internal/core/commands/nhtsa_recalls.go
```diff
@@ -0,0 +1,219 @@
...
+	recallsRepo repositories.DeviceNHTSARecallsRepository
+	ddRepo      repositories.DeviceDefinitionRepository
+	fileURL     *string
+}
+
+func NewSyncNHTSARecallsCommandHandler(dbs func() *db.ReaderWriter, logger *zerolog.Logger, recallsRepo repositories.DeviceNHTSARecallsRepository, ddRepo repositories.DeviceDefinitionRepository, file *string) SyncNHTSARecallsCommandHandler {
+	return SyncNHTSARecallsCommandHandler{dbs: dbs, log: logger, recallsRepo: recallsRepo, ddRepo: ddRepo, fileURL: file}
+}
+
+const NHTSARecallsMatchingVersion = "2022.10.20.0"
```
JamesReate, line 46: lets put a comment here as to what this is used for / what it means

JamesReate (approved):
just some minor things. After merging this let's be sure to test running it in the kubernetes pod. After that we will need to create a new cronjob to run this.

PR #53, opened 2022-10-13
PLA-159 Device definition api upsert make style

rhvivancoeffio:
# Proposed Changes


internal/core/commands/create_ds.go
```diff
@@ -23,10 +24,11 @@ func (*CreateDeviceStyleCommand) Key() string { return "CreateDeviceStyleCommand
 
 type CreateDeviceStyleCommandHandler struct {
 	Repository repositories.DeviceStyleRepository
+	DDCache    services.DeviceDefinitionCacheService
```
JamesReate, line 27: if `DDCache` and `Repository` don't need to be public, just keep them starting with lowercase

PR #58, opened 2022-10-14
EXP-164 Add css icons to grpc response and update proto

0xdev22:
# Proposed Changes
The change adds css icons as part of the response from grpc for features compatibility


pkg/grpc/device_definition.proto
```diff
@@ -250,6 +250,7 @@ message GetDeviceCompatibilityListRequest {
 message Feature {
 	string key = 1;
 	int32 support_level = 2;
+	string css_icon = 3;
```
elffjs, line 253: One thing to consider is that the CSS icon is optional in the database. Should be optional in the message here? We could stick with empty, just a question.

internal/core/queries/device_compatibility.go
```diff
@@ -42,9 +42,12 @@ func (dc GetDeviceCompatibilityQueryHandler) Handle(ctx context.Context, query m
...
 	}
 
-	integFeats := make(map[string]string, len(inf))
+	integFeats := make(map[string]map[string]string, len(inf))
 	for _, k := range inf {
-		integFeats[k.FeatureKey] = k.DisplayName
+		integFeats[k.FeatureKey] = map[string]string{
+			"displayName": k.DisplayName,
+			"cssIcon":     k.CSSIcon.String,
+		}
```
elffjs, line 50: Maybe make a little struct for the value here?

pkg/grpc/device_definition.proto
JamesReate: would that be an optional using something like this? https://pkg.go.dev/github.com/gogo/protobuf/gogoproto#hdr-More_Canonical_Go_Structures

elffjs: Can we leave a comment, at least? "May be blank"?

PR #63, opened 2022-10-20
EXP-157 - Add compat level to proto file and calculate compat level

0xdev22:
# Proposed Changes

- Add level to grpc endpoint response
- Compute compatibility level based on support level
- Add new migration to add feature weights to integration_features table


internal/core/queries/device_compatibility.go
```diff
@@ -9,14 +9,22 @@ import (
 	"github.com/TheFellow/go-mediator/mediator"
 )
 
+const (
+	GOLD_LEVEL   = "Gold"
```
elffjs, line 13: todo: `GoldLevel`, `SilverLevel`, etc.

internal/core/queries/device_compatibility.go
```diff
@@ -39,6 +47,33 @@ func NewGetDeviceCompatibilityQueryHandler(dbs func() *db.ReaderWriter, reposito
...
+func GetDeviceCompatibilityLevel(fd map[string]FeatureDetails, totalWeights float64) string {
+	level := ""
+	total := 0.0
+
+	for _, v := range fd {
+		if v.SupportLevel > 0 {
+			total += v.FeatureWeight
+		} else {
+			total += 0
+		}
```
elffjs, line 59: suggestion: Consider axing this branch.

internal/core/queries/device_compatibility.go
```diff
@@ -39,6 +47,33 @@ func NewGetDeviceCompatibilityQueryHandler(dbs func() *db.ReaderWriter, reposito
...
+	}
+
+	if total != 0 && total < totalWeights {
+		p := (total / totalWeights) * 100
+
+		if p >= 75 {
+			level = GOLD_LEVEL
+		} else if p > 50 && p < 75 {
```
elffjs, line 67: nitpick: `p > 50` would be enough.

pkg/grpc/device_definition.proto
```diff
@@ -268,6 +268,7 @@ message Feature {
 message DeviceCompatibilities {
 	int32 year = 1;
 	repeated Feature features = 2;
+	string level = 3;
```
elffjs, line 271: thought: Any value to using [`enum`](https://developers.google.com/protocol-buffers/docs/proto3#enum) here?

It's attractive because the caller wants one of three values, but maybe it is too fussy.

internal/api/grpc_service.go
```diff
@@ -506,17 +510,28 @@ func (s *GrpcService) GetDeviceCompatibilities(ctx context.Context, in *p_grpc.G
...
+				FeatureWeight: integFeats[f.FeatureKey].FeatureWeight,
+				SupportLevel:  int32(f.SupportLevel),
+			}
+
+			ifeat[f.FeatureKey] = fts
+
 			feats = append(feats, ft)
 		}
 
+		level := queries.GetDeviceCompatibilityLevel(ifeat, totalWeightsCount)
```
elffjs, line 532: question: Is it odd to call something in the `queries` module directly?

internal/core/queries/device_compatibility.go
```diff
@@ -39,6 +47,33 @@ func NewGetDeviceCompatibilityQueryHandler(dbs func() *db.ReaderWriter, reposito
...
+
+	for _, v := range fd {
+		if v.SupportLevel > 0 {
+			total += v.FeatureWeight
+		} else {
+			total += 0
+		}
+	}
+
+	if total != 0 && total < totalWeights {
```
elffjs, line 62: question: Shouldn't `total = totalWeights` be gold?

suggestion: It might be good to warn consumers in the `.proto` that the `level` field might be an empty string.

I agree that "bronze" does not quite describe "has no useful signals whatsoever".

internal/infrastructure/db/migrations/20221006130646_integration_features.sql
```diff
@@ -17,7 +17,7 @@ CREATE UNIQUE INDEX elastic_property_idx ON integration_features (elastic_proper
 INSERT INTO integration_features (feature_key, display_name, elastic_property) VALUES 
     ('ev_battery', 'EV Battery', 'soc'),
     ('battery_voltage', 'Battery Voltage', 'batteryVoltage'),
-    ('fuel_tank', 'Fuel Tank', 'fuelPercentRemaining'),
+    ('fuel_percent_remaining', 'Fuel Remaining', 'fuelPercentRemaining'),
```
elffjs, line 20: chore: We ought to update this on dev and prod.

I can do prod if you do dev.

internal/infrastructure/db/migrations/20221020065955_integration_features_weight.sql
```diff
@@ -0,0 +1,36 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+ALTER TABLE integration_features ADD feature_weight float NULL;
+BEGIN;
```
elffjs, line 6: thought: Goose wraps everything in a transaction, so you may not need to say this.

internal/infrastructure/db/migrations/20221020065955_integration_features_weight.sql
```diff
@@ -0,0 +1,36 @@
...
+    UPDATE integration_features SET feature_weight = 1 WHERE feature_key = 'ev_battery';
```
elffjs, line 7: suggestion (non-blocking): For ease of scanning, something like [this fiddle](https://www.db-fiddle.com/f/8eRDvTna2JQrmJ9dX6tpkL/0) might be nice.

internal/infrastructure/db/migrations/20221020065955_integration_features_weight.sql
```diff
@@ -0,0 +1,36 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+ALTER TABLE integration_features ADD feature_weight float NULL;
```
elffjs, line 5: nitpick: If we write `DROP COLUMN`, why not `ADD COLUMN`?

elffjs (approved):
Extremely minor quibbles from my end, it looks good. We could revisit the calculation when we start using `support_level = 1`.

internal/core/queries/device_compatibility.go
JamesReate: right bronze means there is some degree of compatibility but not a whole lot, eg. cars before 2008 with AutoPi

internal/infrastructure/db/migrations/20221020065955_integration_features_weight.sql
JamesReate: correct, probably not necessary, regardless test without it.

pkg/grpc/device_definition.proto
JamesReate: could be our first enum use case in gRPC protos!

JamesReate (commented):
you can add `//nolint` above those VehicleData / Info lines so that the linter doesn't complain. With our objective of making device definitions more generic, at some point we may get rid of VehicleInfo and just use the attributes, but for now ok just to linter ignore those.

internal/infrastructure/db/migrations/20221020065955_integration_features_weight.sql
0xdev22: its not really a requirement for postgres

elffjs: It's not required, I just like regularity.

PR #65, opened 2022-10-20
EXP-33 Fuel images

Allyson-English:
# Proposed Changes
load fuel api images to table


cmd/device-definitions-api/fuel_image_service.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (fs *FuelServiceAPI) writeToTable(ctx context.Context, data []deviceData, prodID int, prodFormat int) error {
+
+	for _, d := range data {
+		for n := range d.Models {
+
+			img, err := fs.fetchDeviceImage(d.Make, d.Models[n].Model, d.Models[n].Year, prodID, prodFormat)
+			if err != nil {
+				continue
```
JamesReate, line 90: let's add a log that includes the MMY and deviceDefinitionId

cmd/device-definitions-api/fuel_image_service.go
```diff
@@ -0,0 +1,195 @@
...
+	imageReqURL := fmt.Sprintf("/%s?api_key=%s&productID=%d&productFormatIDs=%d", vehicleID, fs.Key, prodID, prodFormat)
+	imageResp, err := http.Get(fs.ImageURL + imageReqURL)
+	if err != nil {
+		return image{}, err
+	}
+	imageData, err := ioutil.ReadAll(imageResp.Body)
+	if err != nil {
+		return image{}, err
+	}
+	imageURL := gjson.Get(string(imageData), "products.0.productFormats.0.assets.0.url").Str
```
JamesReate, line 164: - if there is more than one image, eg. i think it returned multiple colors, should we return an array of images? I understand we may not need it all right now, so if too much to handle just leave a TODO comment
- what happens if there are no images returned in the json body? or would we get a different http status code when that happens?

cmd/device-definitions-api/fuel_image_service.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return image{}, err
+	}
+	vehicleID := gjson.Get(string(vehicleData), "0.id").Str
+	imageReqURL := fmt.Sprintf("/%s?api_key=%s&productID=%d&productFormatIDs=%d", vehicleID, fs.Key, prodID, prodFormat)
+	imageResp, err := http.Get(fs.ImageURL + imageReqURL)
+	if err != nil {
+		return image{}, err
+	}
+	imageData, err := ioutil.ReadAll(imageResp.Body)
```
JamesReate, line 160: do we want to check for the imageResp status code before reading the body? eg. return err if not 200

internal/config/settings.go
```diff
@@ -24,4 +24,7 @@ type Settings struct {
 	DrivlyVINAPIURL                string         `yaml:"DRIVLY_VIN_API_URL"`
 	DrivlyOfferAPIURL              string         `yaml:"DRIVLY_OFFER_API_URL"`
 	Redis                          redis.Settings `yaml:"REDIS"`
+	FuelAPIVehiclesEndpoint        string         `yaml:"FUEL_API_SEARCH_VEHICLES_ENDPOINT"`
```
JamesReate, line 27: I'll add secrets for these to this same PR,

cmd/device-definitions-api/fuel_image_service.go
Allyson-English: added here: https://github.com/DIMO-Network/device-definitions-api/blob/e8a53e6cd46a09a01963906123c180c27b409162/cmd/device-definitions-api/fuel_image_service.go#L162-L180

cmd/device-definitions-api/fuel_image_service.go
Allyson-English: updated to grab array of all image urls returned 
https://github.com/DIMO-Network/device-definitions-api/blob/e8a53e6cd46a09a01963906123c180c27b409162/cmd/device-definitions-api/fuel_image_service.go#L192-L198

Allyson-English: if no images are returned, we'll get a 404 error

PR #70, opened 2022-10-28
Include fuel url in search resp

Allyson-English:
# Proposed Changes
add image url to search endpoint body


internal/core/commands/sync_dd_search_data.go
```diff
@@ -64,16 +64,24 @@ func (ch SyncSearchDataCommandHandler) Handle(ctx context.Context, query mediato
...
-			SearchDisplay: sd,
-			Make:          definition.R.DeviceMake.Name,
-			Model:         definition.Model,
-			Year:          int(definition.Year),
-			SubModels:     sm,
-			ImageURL:      definition.ImageURL.String,
-			MakeSlug:      definition.R.DeviceMake.NameSlug,
-			ModelSlug:     definition.ModelSlug,
+
+		for _, img := range definition.R.Images {
```
JamesReate, line 68: So an issue here is that if the definition has 0 images, it won't get added to the document. Also, if there are more than one images that match the `if` criteria below, you would end up with duplicate models.

What I would recommend is to first set the `docs[i]` from the `definition`, and then loop through the images setting a `thumbnail` variable. After the image loop, then set the `ImageURL` on the `docs[i]` from the image you pulled out of the images array.

PR #81, opened 2022-11-08
PLA-298 get edmunds loader working again use

zavaboy:
# Proposed Changes
Added more support for external_ids
Changes are for loaders

### Impacted Routes
GetDeviceDefinitionBySource
UpdateDeviceDefinition


internal/core/commands/update_dd.go
```diff
@@ -35,6 +36,7 @@ type UpdateDeviceDefinitionCommand struct {
 	DeviceTypeID string `json:"device_type_id"`
 	// DeviceAttributes sets definition metadata eg. vehicle info. Allowed key/values are defined in device_types.properties
 	DeviceAttributes []*coremodels.UpdateDeviceTypeAttribute `json:"deviceAttributes"`
+	ExternalIds      []coremodels.ExternalID                 `json:"externalIds"`
```
JamesReate, line 39: should we use a pointer here same as the property above? `[]*coremodels.ExternalID`

PR #85, opened 2022-11-10
Complete implementation of pagination for compatibility

0xdev22:
# Proposed Changes


JamesReate (approved):
made some minor tweaks, seems like this should work, we'll try it

PR #90, opened 2022-11-14
PLA-406 definition-sync-crash

Allyson-English:
# Proposed Changes
investigating the definition sync crash found that that the `definition.R.DeviceType.Metadatakey` value was returning nil
the proposed solution creates an empty string that is overwritten if the value is not nil, which works locally:
```
		def := *definition.R
		dType := def.DeviceType
		var deviceType string
		if dType != nil {
			deviceType = definition.R.DeviceType.Metadatakey
		}
```

one reason why this solution may not work is that the same value is used as a map key in the section below where the error was initially occurring:
https://github.com/DIMO-Network/device-definitions-api/blob/33dfdad9e00953ac7ff8580c8e5c8a88e3a47c6b/internal/core/commands/sync_dd_search_data.go#L97-L112

other ideas are welcome!


JamesReate:
Investigating a bit, I think `definition.R.DeviceType.Metadatakey` being nil on your local is more of a config issue - in dev and prod this value was set when the jobs ran. What probably does make sense along your line of thought, is to fail with `.fatal` if the definition.R.DeviceType.Metadatakey is not set. So let's update his PR to just do that.

I just ran the script in prod, and it suceeded fine - which makes me think that at the time it had been failing just due to a timeout b/c elastic was busy or something like that.

Allyson-English:
@JamesReate updated with the changes you proposed above!

elffjs:
I think the crash I saw was this: you can reach [this line](https://github.com/DIMO-Network/device-definitions-api/blob/b22e6a444ad263461d04eeade0783009ecccf546/internal/infrastructure/elastic/elastic_app_search_service.go#L296) in `buildAndExecRequest` with `resp` being nil.

internal/infrastructure/elasticsearch/elastic_search_service.go
```diff
@@ -69,9 +69,10 @@ func (d *ElasticSearch) buildAndExecRequest(method, url string, obj interface{},
 	req.Header.Set("Authorization", "ApiKey "+d.Token)
 	var resp *http.Response
 	var err error
```
elffjs, line 71: Does this `err` ever get assigned to before line 106? In particular, it seems like `err == nil` on line 76 will always be true, and `err != nil` will always be false.

internal/infrastructure/elasticsearch/elastic_search_service.go
```diff
@@ -69,9 +69,10 @@ func (d *ElasticSearch) buildAndExecRequest(method, url string, obj interface{},
...
+	var nilRespErr error
 
 	for _, backoff := range backoffSchedule {
-		resp, err = d.httpClient.Do(req) // any error resp should return err per docs
+		resp, nilRespErr = d.httpClient.Do(req) // any error resp should return err per docs
```
elffjs, line 75: Could you explain the name a bit?

internal/infrastructure/elasticsearch/elastic_search_service.go
elffjs: I think the worst outcome here is that we fail a few times due to, say, timeout; we didn't pass in an `outObj`, and we return with a nil error like everything is time.

internal/infrastructure/elasticsearch/elastic_search_service.go
```diff
@@ -101,9 +102,13 @@ func (d *ElasticSearch) buildAndExecRequest(method, url string, obj interface{},
 		time.Sleep(backoff)
 	}
 	if objOut != nil {
-		err = json.NewDecoder(resp.Body).Decode(&objOut)
-		if err != nil {
-			return nil, errors.Wrap(err, "error decoding response json")
+		if resp != nil {
+			err = json.NewDecoder(resp.Body).Decode(&objOut)
```
elffjs, line 106: Hm, is it the caller's responsibility to close the body? It doesn't seem like any of the callers are.

internal/core/commands/sync_dd_search_data.go
```diff
@@ -75,6 +75,13 @@ func (ch SyncSearchDataCommandHandler) Handle(ctx context.Context, query mediato
 			}
 		}
 
+		def := *definition.R
+		dType := def.DeviceType
+		var deviceType string
```
elffjs, line 80: Do we want to assign to this? `deviceType := dType.Metadatakey`?

internal/core/commands/sync_dd_search_data.go
```diff
@@ -75,6 +75,13 @@ func (ch SyncSearchDataCommandHandler) Handle(ctx context.Context, query mediato
...
+		if dType == nil {
```
elffjs, line 81: It looks like `device_definitions.device_type_id` is still nullable in the database. I wish it weren't, but given this I wouldn't want to tear down the app if it's missing.

JamesReate: ah good point, making a PR to update `device_type_id` to be not null.

elffjs:
@JamesReate We gave rewriting this using `shared` a shot. I think the functionality is the same, would you mind taking a look?

JamesReate (approved):

PR #95, opened 2022-11-16
EXP-253 - Device definition api reviews

rhvivancoeffio:
# Proposed Changes
Create reviews api


JamesReate (approved):
seems alright to me

PR #97, opened 2022-11-18
EXP-253 - Device definition api review dynamic filter

rhvivancoeffio:
# Proposed Changes
Added Dynamic Filter to reviews


internal/api/grpc_reviews_service.go
```diff
@@ -30,6 +30,21 @@ func (s *GrpcReviewsService) GetReviewsByDeviceDefinitionID(ctx context.Context,
...
 
+func (s *GrpcReviewsService) GetReviews(ctx context.Context, in *p_grpc.GetReviewFilterRequest) (*p_grpc.GetReviewsResponse, error) {
+	qryResult, _ := s.Mediator.Send(ctx, &queries.GetReviewsDynamicFilterQuery{
+		MakeID:             in.MakeId,
+		Model:              in.Model,
+		Year:               int(in.Year),
+		DeviceDefinitionID: in.DeviceDefinitionId,
+		PageIndex:          int(in.PageIndex),
+		PageSize:           int(in.PageSize),
+	})
```
JamesReate, line 41: i wonder if we really need to do a projection to another object here or just pass `in`

internal/core/queries/get_reviews_by_id.go
```diff
@@ -53,6 +57,7 @@ func (qh GetReviewsByIDQueryHandler) Handle(ctx context.Context, query mediator.
 		Comments:           review.Comments,
 		Approved:           review.Approved,
 		ApprovedBy:         review.ApprovedBy,
+		Name:               fmt.Sprintf("%d %s %s", review.R.DeviceDefinition.Year, review.R.DeviceDefinition.R.DeviceMake.Name, review.R.DeviceDefinition.Model),
```
JamesReate, line 60: refactor how name is built with above - since we do same thing in two spots

internal/core/queries/get_reviews_dynamic_filter.go
```diff
@@ -0,0 +1,105 @@
...
+		result.Reviews = append(result.Reviews, &p_grpc.DeviceReview{
+			Id:                 review.ID,
+			Url:                review.URL,
+			ImageURL:           review.ImageURL,
+			Channel:            review.Channel.String,
+			DeviceDefinitionId: review.DeviceDefinitionID,
+			Comments:           review.Comments,
+			Approved:           review.Approved,
+			ApprovedBy:         review.ApprovedBy,
+			Name:               fmt.Sprintf("%d %s %s", review.R.DeviceDefinition.Year, review.R.DeviceDefinition.R.DeviceMake.Name, review.R.DeviceDefinition.Model),
```
JamesReate, line 100: refactor with above too

pkg/grpc/reviews.proto
```diff
@@ -55,8 +56,19 @@ message DeleteReviewRequest {
...
+  int32 year = 2;
+  string model = 3;
+  string  device_definition_id = 4;
+  int32 page_index = 5;
+  int32 page_size = 6;
+}
+
 service ReviewsService {
   rpc GetReviewsByDeviceDefinitionID(GetReviewsByDeviceDefinitionRequest) returns (GetReviewsResponse);
+  rpc GetReviews(GetReviewFilterRequest) returns (GetReviewsResponse);
```
JamesReate, line 71: add comment above that currently being used for admin tool

internal/core/queries/get_reviews_by_id.go
rhvivancoeffio: I created an util function => BuildDeviceDefinitionName

PR #104, opened 2022-11-22
[EXP-276] add metadata to device makes model

Asfiroth:
# Proposed Changes


JamesReate (commented):
Wouldn't this also need a change to the device make proto to expose this over gRPC?

JamesReate (approved):
i think it looks good

PR #112, opened 2022-11-28
PLA-473 - Get make by token id

rhvivancoeffio:
# Proposed Changes
Get make by token id


internal/core/queries/get_dm_by_token_id.go
```diff
@@ -0,0 +1,79 @@
...
+		ID:               v.ID,
+		Name:             v.Name,
+		LogoURL:          v.LogoURL,
+		OemPlatformName:  v.OemPlatformName,
+		NameSlug:         v.NameSlug,
+		ExternalIds:      common.JSONOrDefault(v.ExternalIds),
+		ExternalIdsTyped: common.BuildExternalIds(v.ExternalIds),
+		Metadata:         common.JSONOrDefault(v.Metadata),
+		MetadataTyped:    common.BuildDeviceMakeMetadata(v.Metadata),
+	}
```
JamesReate, line 72: since this isn't needed as a rest endpoint, do we need this projection or could we just return grpc object directly

JamesReate (approved):
Just one comment re: might be able to save on a struct  projection, but lgtm

PR #117, opened 2022-12-09
endpoint to return all mmys

Allyson-English:
# Proposed Changes
add endpoint to return all device make/ model/ years


internal/core/queries/get_dmmy_all.go
```diff
@@ -0,0 +1,45 @@
...
+	all, err := ch.Repository.GetDeviceMMY(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	response := make([]interface{}, 0)
+
+	for _, v := range all {
+		resp := &grpc.DeviceType{
+			Make:  v.R.DeviceMake.Name,
```
JamesReate, line 37: let's use the make slug

internal/infrastructure/db/repositories/device_definition_repo.go
```diff
@@ -123,6 +124,24 @@ func (r *deviceDefinitionRepository) GetAll(ctx context.Context) ([]*models.Devi
 	return dd, err
 }
 
+func (r *deviceDefinitionRepository) GetDeviceMMY(ctx context.Context) ([]*models.DeviceDefinition, error) {
```
JamesReate, line 127: can we rename to `GetAllDevicesMMY`

internal/infrastructure/db/repositories/device_definition_repo.go
```diff
@@ -123,6 +124,24 @@ func (r *deviceDefinitionRepository) GetAll(ctx context.Context) ([]*models.Devi
...
+
+	dd, err := models.DeviceDefinitions(
```
JamesReate, line 129: to make this query more efficient, b/c it is pulling lots of data, let's only select the columns we care about. To do that, [here is an example](https://github.com/volatiletech/sqlboiler#select)

internal/infrastructure/db/repositories/device_definition_repo.go
```diff
@@ -31,6 +31,7 @@ type DeviceDefinitionRepository interface {
 	GetByMakeModelAndYears(ctx context.Context, make string, model string, year int, loadIntegrations bool) (*models.DeviceDefinition, error)
 	GetBySlugAndYear(ctx context.Context, slug string, year int, loadIntegrations bool) (*models.DeviceDefinition, error)
 	GetAll(ctx context.Context) ([]*models.DeviceDefinition, error)
+	GetDeviceMMY(ctx context.Context) ([]*models.DeviceDefinition, error)
```
JamesReate, line 34: you'll need to regenerate the mocks per go generate command towards top of this file. this should handle some of the linter errors

internal/infrastructure/db/repositories/device_definition_repo.go
Allyson-English: I'm having trouble testing that this query returns what I want it to-- can you take a look at it?

PR #120, opened 2022-12-13
PLA-645 - Device definition api hardware template

rhvivancoeffio:
# Proposed Changes
Device definition api hardware template


internal/core/commands/create_dm_test.go
```diff
@@ -46,16 +46,18 @@ func (s *CreateDeviceMakeCommandHandlerSuite) TestCreateDeviceMakeCommand_Succes
 	ctx := context.Background()
 
 	name := "Ford"
+	templateId := "01"
 
 	dm := &models.DeviceMake{
 		ID:   "1",
```
JamesReate, line 52:
```suggestion
		ID:   ksuid.New().String(),
```

internal/core/common/utils.go
```diff
@@ -159,21 +159,23 @@ func BuildFromDeviceDefinitionToQueryResult(dd *repoModel.DeviceDefinition) (*mo
...
-			Year:      int(dd.Year),
-			MakeSlug:  dd.R.DeviceMake.NameSlug,
-			ModelSlug: dd.ModelSlug,
+			Type:               strings.TrimSpace(dd.R.DeviceType.ID),
+			Make:               dd.R.DeviceMake.Name,
+			Model:              dd.Model,
+			Year:               int(dd.Year),
+			MakeSlug:           dd.R.DeviceMake.NameSlug,
+			ModelSlug:          dd.ModelSlug,
+			HardwareTemplateID: dd.HardwareTemplateID.String,
```
JamesReate, line 178: I don't think DeviceType will need the hardware template id

internal/core/models/device_definition_models.go
```diff
@@ -106,3 +109,7 @@ type ExternalID struct {
 type DeviceMakeMetadata struct {
 	RideGuideLink string `json:"ride_guide_link"`
 }
+
+type GetDeviceDefinitionHardwareTemplateQueryResult struct {
```
JamesReate, line 113: what would this be used for?

PR #124, opened 2022-12-19
All devices mmy

Allyson-English:
# Proposed Changes
add endpoint that returns all devices mmy to be used to generate paths


JamesReate (approved):
tested locally with my prod equivalent DB and it works nicely!

PR #130, opened 2023-01-22
EXP-332 - Recalls gRPC endpoint changed to Streaming

rhvivancoeffio:
# Proposed Changes


JamesReate (approved):
lgtm, let's just deprecate the old, non stream method

PR #137, opened 2023-02-17
PLA-1007 - Sync VIN Decoding

rhvivancoeffio:
# Proposed Changes


internal/core/commands/sync_vin_numbers.go
```diff
@@ -0,0 +1,168 @@
...
+			Logger()
+
+		vinDecodeNumber, err := models.FindVinNumber(ctx, dc.dbs().Reader, vin.String())
+		if err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				localLog.Debug().Msg("No rows to decode vin numbers from db")
+			}
+		}
+
+		if vinDecodeNumber == nil {
```
JamesReate, line 79: `if vinDecodeNumber != nil ` we should return the object and short circuit the rest of the operation since we already have seen it before.

JamesReate: oh i see it just continues

internal/core/commands/sync_vin_numbers.go
```diff
@@ -0,0 +1,168 @@
...
+					models.DeviceStyleWhere.SubModel.EQ(vinInfo.SubModel),
+					models.DeviceStyleWhere.Name.EQ(vinInfo.StyleName)).One(ctx, dc.dbs().Reader)
+				if errors.Is(err, sql.ErrNoRows) {
+					// insert
+					style = &models.DeviceStyle{
+						ID:                 ksuid.New().String(),
+						DeviceDefinitionID: dd.ID,
+						Name:               vinInfo.StyleName,
+						ExternalStyleID:    common.SlugString(vinInfo.StyleName),
+						Source:             "drivly",
```
JamesReate, line 137: source should be from vinInfo.Source?

internal/core/services/vin_decoding_service.go
```diff
@@ -0,0 +1,117 @@
...
+				Str("vin", vin).
+				Msg("failed to decode vin from vincario")
+		}
+
+		if vinVincarioInfo != nil {
+			result.VIN = vinVincarioInfo.VIN
+			result.Year = strconv.Itoa(vinVincarioInfo.ModelYear)
+			result.Make = vinVincarioInfo.Make
+			result.Model = vinVincarioInfo.Model
+			result.Source = "vincario"
```
JamesReate, line 58: i think there should also be a `vincarioInfo.GetStyle()`

internal/core/models/vin_decoding_models.go
```diff
@@ -0,0 +1,14 @@
+package models
+
+import "github.com/volatiletech/null/v8"
+
+type VINDecodingInfoData struct {
+	VIN       string
+	Make      string
+	Model     string
+	SubModel  string
```
JamesReate, line 9: is `SubModel` being used? Seems the same as `StyleName`

JamesReate (approved):

PR #138, opened 2023-02-20
[PLA-1009] bulk decode vins

Asfiroth:
# Proposed Changes
- add bulk decode api


JamesReate (approved):
i think it's good ready to try it out
dont forget to regenerate the swagger docs!

PR #140, opened 2023-02-23
decode_provider added to vin_numbers

rhvivancoeffio:
# Proposed Changes

### DIFF
--- a/internal/core/queries/decode_vin.go
+++ b/internal/core/queries/decode_vin.go
@@ -94,9 +94,18 @@ func (dc DecodeVINQueryHandler) Handle(ctx context.Context, query mediator.Messa
 		return resp, nil
 	}
 
-	if len(vinInfo.StyleName) == 0 {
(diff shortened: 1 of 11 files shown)


internal/core/queries/decode_vin.go
```diff
@@ -90,7 +90,12 @@ func (dc DecodeVINQueryHandler) Handle(ctx context.Context, query mediator.Messa
 
 	vinInfo, err := dc.vinDecodingService.GetVIN(vin.String(), dt)
 	if err != nil {
-		localLog.Err(err).Msg("failed to decode vin from drivly or vincario")
+		localLog.Err(err).Msgf("failed to decode vin from %s", vinInfo.Source)
+		return resp, nil
+	}
+
+	if len(vinInfo.StyleName) == 0 {
+		localLog.Err(err).Msgf("vin %s stylename is empty", vinInfo.VIN)
```
JamesReate, line 98: just change to Warn - not a big deal if we don't have the style

PR #151, opened 2023-03-15
[PLA-1019] add grpc endpoint for dd sync

Asfiroth:
# Proposed Changes
- ddsync grpc endpoint
-


internal/api/grpc_definitions_service.go
```diff
@@ -760,3 +760,13 @@ func (s *GrpcDefinitionsService) GetDeviceDefinitionsWithHardwareTemplate(ctx co
 
 	return qryResult.(*p_grpc.GetDevicesMMYResponse), nil
 }
+
+func (s *GrpcDefinitionsService) SyncDeviceDefinitionsWithElasticSearch(ctx context.Context, in *emptypb.Empty) (*p_grpc.SyncStatusResult, error) {
+	command := &commands.SyncSearchDataCommand{}
+
+	commandResult, _ := s.Mediator.Send(ctx, command)
```
JamesReate, line 767: this looks like it would block - I guess we can try it out as is and then see if needs to be wrapped in a go func()

JamesReate (approved):

PR #163, opened 2023-04-20
[PLA-1084] add metadata to dd from vincario

Asfiroth:
# Proposed Changes
- add metadata to dd from vincario

### DIFF
--- a/internal/infrastructure/gateways/vincario_api_service.go
+++ b/internal/infrastructure/gateways/vincario_api_service.go
@@ -212,7 +212,7 @@ func (v *VincarioInfoResponse) GetMetadata() null.JSON {
 		"mpg_highway":            nil,
 		"mpg_city":               nil,
 		"fuel_tank_capacity_gal": nil,
-		"mpg":                    v.EngineDisplacement,
+		"mpg":                    nil,
 	}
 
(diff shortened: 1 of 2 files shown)


internal/infrastructure/gateways/vincario_api_service.go
```diff
@@ -196,3 +198,28 @@ func (v *VincarioInfoResponse) GetStyle() string {
...
+		"fuel_type":              v.FuelType,
+		"driven_wheels":          v.Drive,
+		"number_of_doors":        v.NumberOfDoors,
+		"base_msrp":              nil,
+		"epa_class":              v.EmissionStandard,
+		"vehicle_type":           v.Body,
+		"mpg_highway":            nil,
+		"mpg_city":               nil,
+		"fuel_tank_capacity_gal": nil,
+		"mpg":                    v.EngineDisplacement,
```
JamesReate, line 215: this one set to nil, i don't think vincario has this. Maybe fuel consumption?

Asfiroth: vincario does not have a fuel consumption property so I'm setting this to nil.

JamesReate (approved):

PR #165, opened 2023-05-02
PLA-1343 - Device definitions api grpc middleware

rhvivancoeffio:
# Proposed Changes


internal/api/api.go
```diff
@@ -195,3 +199,19 @@ func startMonitoringServer(logger zerolog.Logger) {
 
 	logger.Info().Str("port", "8888").Msg("Started monitoring web server.")
 }
+
+func httpMetricsMiddleware(c *fiber.Ctx) error {
```
JamesReate, line 203: only thing to be aware of is will this catch the path with the parameter or just the path

JamesReate (approved):

PR #166, opened 2023-05-08
Add tokenID to Integrations proto and column to integrations table

0xdev22:
# Proposed Changes


internal/infrastructure/db/models/boil_queries.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by SQLBoiler 4.14.1 (https://github.com/volatiletech/sqlboiler). DO NOT EDIT.
```
elffjs, line 1: Can you get your SQLBoiler and protoc up-to-date? Good to not go backwards if we don't have to.

internal/infrastructure/db/migrations/20230508160305_add_token_id_integrations.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
```
elffjs, line 3:
```suggestion
```
Here and below.

internal/infrastructure/db/migrations/20230508160305_add_token_id_integrations.sql
```diff
@@ -0,0 +1,13 @@
...
+SET search_path = device_definitions_api, public;
+ALTER TABLE integrations ADD token_id numeric(78,0) null;
```
elffjs, line 4: Minor points:

* I wouldn't mind this being an `int`. I don't think there will be too many.
* For some reason I find `ADD COLUMN` comforting to see, even though the `COLUMN` is optional.
* Contradicting myself immediately: I would drop the last `null`.

elffjs (commented):
Do we do anything with this new column? I would have expected changes to [`grpc_definitions_service.go`](https://github.com/DIMO-Network/device-definitions-api/blob/main/internal/api/grpc_definitions_service.go).

internal/infrastructure/db/migrations/20230508160305_add_token_id_integrations.sql
0xdev22: I thought the field was meant to be nullable @elffjs

elffjs: It should be nullable, but that's the default. It's not wrong!

internal/infrastructure/db/models/integrations.go
```diff
@@ -31,8 +32,9 @@ type Integration struct {
 	CreatedAt time.Time `boil:"created_at" json:"created_at" toml:"created_at" yaml:"created_at"`
 	UpdatedAt time.Time `boil:"updated_at" json:"updated_at" toml:"updated_at" yaml:"updated_at"`
 	// How often can integration be called in seconds
-	RefreshLimitSecs int       `boil:"refresh_limit_secs" json:"refresh_limit_secs" toml:"refresh_limit_secs" yaml:"refresh_limit_secs"`
-	Metadata         null.JSON `boil:"metadata" json:"metadata,omitempty" toml:"metadata" yaml:"metadata,omitempty"`
+	RefreshLimitSecs int               `boil:"refresh_limit_secs" json:"refresh_limit_secs" toml:"refresh_limit_secs" yaml:"refresh_limit_secs"`
+	Metadata         null.JSON         `boil:"metadata" json:"metadata,omitempty" toml:"metadata" yaml:"metadata,omitempty"`
+	TokenID          types.NullDecimal `boil:"token_id" json:"token_id,omitempty" toml:"token_id" yaml:"token_id,omitempty"`
```
elffjs, line 37: Why is this still NullDecimal? That seems suspect.

internal/core/models/integration_models.go
```diff
@@ -49,4 +49,5 @@ type GetIntegrationQueryResult struct {
 	AutoPiDefaultTemplateID      int                    `json:"auto_pi_default_template_id"`
 	RefreshLimitSecs             int                    `json:"refresh_limit_secs"`
 	AutoPiPowertrainToTemplateID map[PowertrainType]int `json:"auto_pi_power_train_to_template_id,omitempty"`
+	TokenID                      int                    `json:"token_id"`
```
elffjs, line 52: What populates this? Is there some reflection going on? How easy is it to run this locally?

internal/infrastructure/db/migrations/20230508160305_add_token_id_integrations.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+-- +goose StatementBegin
+SET search_path = device_definitions_api, public;
+ALTER TABLE integrations ADD COLUMN token_id int;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+SELECT 'down SQL query';
```
elffjs, line 9:
```suggestion
```

internal/core/models/integration_models.go
elffjs: Like I think there are `get_in_*.go` and associated tests here that probably have to be dealt with:

https://github.com/DIMO-Network/device-definitions-api/tree/40d1c99991bec422a39252e3045e76c5f3d89c64/internal/core/queries

internal/core/queries/get_in_all.go
```diff
@@ -53,6 +53,7 @@ func (ch GetAllIntegrationQueryHandler) Handle(ctx context.Context, _ mediator.M
 			Vendor:                  v.Vendor,
 			AutoPiDefaultTemplateID: im.AutoPiDefaultTemplateID,
 			RefreshLimitSecs:        v.RefreshLimitSecs,
+			TokenID:                 v.TokenID.Int,
```
JamesReate, line 56: is it ok for TokenID to be `0` if it is null in the db? Or are we not expecting these to be null eventually?

JamesReate (commented):
I just left a question and plus 1 on having a comment regarding `token_id`

pkg/grpc/device_definition.proto
```diff
@@ -143,6 +143,7 @@ message Integration {
   int32 auto_pi_default_template_id = 5;
   AutoPiPowertrainTemplate auto_pi_powertrain_template = 6;
   int32 refresh_limit_secs = 7;
+  uint64 token_id = 8;  // token_id can have a 0 value, which means it has not yet been minted and no token id has been assigned. This case should be checked for and handled.
```
elffjs, line 146: Does this show up in the generated source? My side goal here is that if we look at the field in an IDE it will show this hint.

0xdev22: Yes it does

elffjs: That's good. Can I be picky and ask for it above the variable? That seems more standard, e.g.,

https://github.com/googleapis/googleapis/blob/master/google/cloud/bigquery/storage/v1beta2/protobuf.proto

internal/infrastructure/db/migrations/20230508160305_add_token_id_integrations.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+SET search_path = device_definitions_api, public;
+ALTER TABLE integrations ADD COLUMN token_id int;
```
elffjs, line 4: I can't believe I forgot—unique?

internal/core/queries/get_in_all.go
elffjs: @JamesReate Is that good enough for you?

internal/core/commands/create_in.go
```diff
@@ -42,6 +44,7 @@ func (ch CreateIntegrationCommandHandler) Handle(ctx context.Context, query medi
 	i.Vendor = command.Vendor
 	i.Type = command.Type
 	i.Style = command.Style
+	i.TokenID = null.IntFrom(command.TokenID)
```
elffjs, line 47: I would make one exception here: if the input is `0` then leave this as null. If this seems like too much of a headache then just drop the nullable, drop the unique, make the default 0.

elffjs (approved):
I'd wait for James to merge, but it looks good to me!

PR #168, opened 2023-05-10
Add Grpc endpoint to return integration by tokenID

0xdev22:
# Proposed Changes


elffjs (changes requested):
Can we combine more with the other method? Preparing the response looks the same whether you query by id or token. Looks good otherwise.

It’s interesting that most of these handlers ignore the error return value from the query. Not your fault.

internal/api/grpc_definitions_service.go
```diff
@@ -185,6 +185,31 @@ func (s *GrpcDefinitionsService) GetIntegrationByID(ctx context.Context, _ *p_gr
 	return result, nil
 }
 
+func (s *GrpcDefinitionsService) GetIntegrationByTokenID(ctx context.Context, in *p_grpc.GetIntegrationByTokenIDRequest) (*p_grpc.Integration, error) {
+	qryResult, _ := s.Mediator.Send(ctx, &queries.GetIntegrationByTokenIDQuery{
+		TokenID: int(in.TokenId),
+	})
+
+	item := qryResult.(coremodels.GetIntegrationQueryResult)
```
elffjs, line 193: Like here onwards.

internal/core/queries/get_in_by_token_id.go
```diff
@@ -0,0 +1,85 @@
...
+			return nil, &exceptions.NotFoundError{
+				Err: fmt.Errorf("could not find integration with tokenID: %d ", qry.TokenID),
+			}
+		}
+		return nil, &exceptions.InternalError{
+			Err: fmt.Errorf("failed to get integration using tokenID: %d", qry.TokenID),
+		}
+	}
+
+	response := coremodels.GetIntegrationQueryResult{}
```
elffjs, line 58: And here onwards.

internal/core/queries/get_in_by_token_id.go
```diff
@@ -0,0 +1,85 @@
...
+		Vendor:                  integration.Vendor,
+		AutoPiDefaultTemplateID: im.AutoPiDefaultTemplateID,
+		RefreshLimitSecs:        integration.RefreshLimitSecs,
+		TokenID:                 integration.TokenID.Int,
+	}
+
+	if im.AutoPiPowertrainToTemplateID != nil {
+		response.AutoPiPowertrainToTemplateID = im.AutoPiPowertrainToTemplateID
+	}
+	log.Println(response, "responseresponse")
```
elffjs, line 83: Can we drop this before merging?

elffjs (approved):

PR #172, opened 2023-06-08
PRO-1413 - use transaction for vin decode

JamesReate:
# Proposed Changes


internal/core/queries/decode_vin.go
```diff
@@ -89,8 +89,13 @@ func (dc DecodeVINQueryHandler) Handle(ctx context.Context, query mediator.Messa
 	)
 
 	metrics.Success.With(prometheus.Labels{"method": VinRequests}).Inc()
+	tx, err := dc.dbs().Writer.BeginTx(ctx, nil)
```
elffjs, line 92: You might want serializable, I don't know. Is read committed enough for you?

JamesReate: yeah, left that while i did some research. 
in agreement, ended up going for serializable, thanks for validating

PR #175, opened 2023-08-09
PRO-1706 - Device definition api powertraintype sync

rhvivancoeffio:
# Proposed Changes

### DIFF
--- a/internal/core/commands/sync_dd_power_train_type.go
+++ b/internal/core/commands/sync_dd_power_train_type.go
@@ -37,6 +37,8 @@ func NewSyncPowerTrainTypeCommandHandler(dbs func() *db.ReaderWriter, logger zer
 
 func (ch SyncPowerTrainTypeCommandHandler) Handle(ctx context.Context, _ mediator.Message) (interface{}, error) {
 
+	const powerTrainType = "powertrain_type"
+
 	all, err := models.DeviceDefinitions(models.DeviceDefinitionWhere.Verified.EQ(true),
 		models.DeviceDefinitionWhere.DeviceTypeID.EQ(null.StringFrom("vehicle")),
 		qm.Load(models.DeviceDefinitionRels.DeviceStyles),
@@ -97,9 +99,10 @@ func (ch SyncPowerTrainTypeCommandHandler) Handle(ctx context.Context, _ mediato
 		}
 
 		for key, _ := range metadataAttributes[metadataKey].(map[string]any) {
-			if key == "powertrain_type" {
-				powerTrainType := ch.resolvePowerTrainTypeByMake(ctx, powerTrainTypeData, definition)
-				metadataAttributes[metadataKey].(map[string]interface{})["powertrain_type"] = powerTrainType
+
+			if key == powerTrainType {
+				powerTrainTypeValue := ch.resolvePowerTrainTypeByMake(ctx, powerTrainTypeData, definition)
+				metadataAttributes[metadataKey].(map[string]interface{})[powerTrainType] = powerTrainTypeValue
 			}
 		}
 
--- a/internal/core/commands/sync_dd_power_train_type.go
+++ b/internal/core/commands/sync_dd_power_train_type.go
@@ -148,9 +148,7 @@ func (ch SyncPowerTrainTypeCommandHandler) resolvePowerTrainTypeByMake(ctx conte
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


internal/core/commands/sync_dd_power_train_type.go
```diff
@@ -0,0 +1,201 @@
...
+func NewSyncPowerTrainTypeCommandHandler(dbs func() *db.ReaderWriter, logger zerolog.Logger) SyncPowerTrainTypeCommandHandler {
+	return SyncPowerTrainTypeCommandHandler{DBS: dbs, logger: logger}
+}
+
+func (ch SyncPowerTrainTypeCommandHandler) Handle(ctx context.Context, _ mediator.Message) (interface{}, error) {
+
+	all, err := models.DeviceDefinitions(models.DeviceDefinitionWhere.Verified.EQ(true),
+		qm.Load(models.DeviceDefinitionRels.DeviceStyles),
+		qm.Load(models.DeviceDefinitionRels.DeviceType),
+		qm.Load(models.DeviceDefinitionRels.DeviceMake)).All(ctx, ch.DBS().Reader)
```
JamesReate, line 42: add a where the device_type is vehicle etc. Maybe query the device_types table and find the one that has powertrain_type as an attribute and pick that key?

internal/core/commands/sync_dd_power_train_type.go
```diff
@@ -0,0 +1,201 @@
...
+				for _, deviceAttribute := range deviceTypeAttributes["properties"] {
+					metaData[deviceAttribute.Name] = deviceAttribute.DefaultValue
+				}
+			}
+
+			metadataAttributes[metadataKey] = metaData
+		}
+
+		for key, _ := range metadataAttributes[metadataKey].(map[string]any) {
+			if key == "powertrain_type" {
```
JamesReate, line 98: let's make this a `const` defined towards beginning of file

internal/core/commands/sync_dd_power_train_type.go
```diff
@@ -0,0 +1,201 @@
...
+				powerTrainType := ch.resolvePowerTrainTypeByMake(ctx, powerTrainTypeData, definition)
+				metadataAttributes[metadataKey].(map[string]interface{})["powertrain_type"] = powerTrainType
```
JamesReate, line 100: Let's add a log message any time that the powertrain_type was previously set and we're changing it. Eg. "Updating powertrain_type from ICE to HEV."

Also, if the powertrain_type is unchanged, should we just skip to save the DB upsert call?

internal/core/commands/sync_dd_power_train_type.go
```diff
@@ -0,0 +1,201 @@
...
+
+		if err = definition.Upsert(ctx, ch.DBS().Writer, true, []string{models.DeviceDefinitionColumns.ID}, boil.Infer(), boil.Infer()); err != nil {
+			return nil, err
+		}
+	}
+
+	return SyncPowerTrainTypeCommandResult{true}, nil
+}
+
+func (ch SyncPowerTrainTypeCommandHandler) resolvePowerTrainTypeByMake(ctx context.Context, powerTrainTypeData coremodels.PowerTrainTypeRuleData,
```
JamesReate, line 117:
```suggestion
func (ch SyncPowerTrainTypeCommandHandler) resolvePowerTrainType(ctx context.Context, powerTrainTypeData coremodels.PowerTrainTypeRuleData,
```
Consider renaming since this function does more than just resolving By Make

internal/core/commands/sync_dd_power_train_type.go
```diff
@@ -0,0 +1,201 @@
...
+						return ptType.Type
+					}
+				}
+
+			}
+		}
+	}
+
+	// Default
+	defaultPowerTrainType := ""
```
JamesReate, line 138: should this be set to `"ICE"` ?

internal/core/commands/sync_dd_power_train_type.go
```diff
@@ -0,0 +1,201 @@
...
+						return item.Type
+					}
+				}
+			}
+		}
+
+	}
+
+	// Resolve Vincario Data
+	ch.logger.Info().Msg("Validating Vincario Data")
```
JamesReate, line 186:
```suggestion
	ch.logger.Info().Msg("Looking up PowerTrain from Vincario Data")
```

JamesReate (approved):

PR #180, opened 2023-08-23
PRO-1785 - Return Default DD

rhvivancoeffio:
# Proposed Changes


internal/core/services/vin_decoding_service.go
```diff
@@ -178,3 +196,23 @@ func buildFromDrivly(info *gateways.DrivlyVINResponse) *models.VINDecodingInfoDa
 func buildDrivlyStyleName(vinInfo *gateways.DrivlyVINResponse) string {
 	return strings.TrimSpace(vinInfo.Trim + " " + vinInfo.SubModel)
 }
+
+func buildFromDD(vin string, info *repoModel.DeviceDefinition) *models.VINDecodingInfoData {
```
JamesReate, line 200: is there a way we could fill in the metadata - specifically what we depend on for PowerTrain?

JamesReate: maybe just set the metadata to the metadata from the database

JamesReate (approved):

PR #181, opened 2023-08-24
PRO-1878 - add device attribute to device style

rhvivancoeffio:
# Proposed Changes

### DIFF
--- a/internal/core/queries/get_ds_by_id.go
+++ b/internal/core/queries/get_ds_by_id.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"github.com/DIMO-Network/device-definitions-api/internal/core/common"
 	"github.com/DIMO-Network/device-definitions-api/internal/core/services"
+	"strings"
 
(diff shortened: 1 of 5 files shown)


internal/core/queries/get_ds_by_id.go
```diff
@@ -57,5 +59,16 @@ func (ch GetDeviceStyleByIDQueryHandler) Handle(ctx context.Context, query media
...
 	}
 
+	dd, err := ch.DDCache.GetDeviceDefinitionByID(ctx, result.DeviceDefinitionID)
+	if err != nil {
+		return nil, &exceptions.InternalError{
+			Err: fmt.Errorf("failed to get device definition"),
+		}
+	}
+
+	result.DeviceDefinition = coremodels.GetDeviceDefinitionStyleQueryResult{
```
JamesReate, line 69: we'll need to find the powertrain attribute and modify it if the device style name matches any of the hybrid / plugin hybrid names

JamesReate:
after discussion, found need to have a check to no override powertrain that comes from DD if powertrainType is empty

JamesReate (approved):

PR #182, opened 2023-08-29
PRO-1898 - Device definitions api custom mediator

rhvivancoeffio:
# Proposed Changes


JamesReate (approved):

internal/core/commands/demo_command.go
```diff
@@ -0,0 +1,37 @@
+package commands
```
JamesReate, line 1: let's remove this one once done

JamesReate (approved):

PR #189, opened 2023-10-12
SI-2110-update-integrations-tbl

Allyson-English:
[ticket](https://linear.app/dimo/issue/SI-2110/add-points-and-manufacturer-link-to-integrations-table)

- add `points` and `manufacturer_token_id` to integrations table and grpc integrations endpoint (optional fields)

### Notes
- this does not update the integration creation commands so `points` and `manufacturer_token_id` will have to be added manually (they are optional fields so this will not be a problem if new integrations are added using these commands in the future); making these fields required and updating commands and/ or adding a command to insert these values might be a future TODO


pkg/grpc/device_definition.proto
```diff
@@ -151,6 +151,8 @@ message Integration {
     int32 ICE = 3;
     int32 PHEV = 4;
   }
+  optional int64 points = 9;
+  optional uint64 manufacturer_token_id = 10;
```
elffjs, line 155: It looks like in this neck of the woods they don't do optional. Maybe that's even correct. Let's take it off.

pkg/grpc/device_definition.proto
```diff
@@ -151,6 +151,8 @@ message Integration {
     int32 ICE = 3;
     int32 PHEV = 4;
   }
+  optional int64 points = 9;
```
elffjs, line 154: Since we're requiring this in the DB, I have no qualms about "requiring" it here too.

internal/infrastructure/db/migrations/20231012114159_update_integrations_tbl.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+SET search_path = device_definitions_api, public;
+ALTER TABLE integrations ADD COLUMN points int;
```
elffjs, line 4: Let's `NOT NULL` here and default to `0`.

internal/infrastructure/db/migrations/20231012114159_update_integrations_tbl.sql
```diff
@@ -0,0 +1,15 @@
...
+ALTER TABLE integrations ADD COLUMN manufacturer_token_id numeric(78);
```
elffjs, line 5: Can we foreign key this to `device_makes`?

internal/core/models/integration_models.go
```diff
@@ -50,4 +50,6 @@ type GetIntegrationQueryResult struct {
 	RefreshLimitSecs             int                    `json:"refresh_limit_secs"`
 	AutoPiPowertrainToTemplateID map[PowertrainType]int `json:"auto_pi_power_train_to_template_id,omitempty"`
 	TokenID                      int                    `json:"token_id"`
+	Points                       *int64                 `json:"points"`
+	ManufacturerTokenID          *uint64                `json:"manufacturer_token_id"`
```
elffjs, line 54: I guess the pattern is that these are `int`.

internal/core/queries/get_dd_by_id_with_rels.go
```diff
@@ -62,6 +64,19 @@ func (ch GetDeviceDefinitionWithRelsQueryHandler) Handle(ctx context.Context, qu
 				Region:       di.Region,
 				Capabilities: common.JSONOrDefault(di.Features),
 			})
+
```
elffjs, line 67: Well, this doesn't have `tokenId` either. I almost want to leave this alone.

internal/core/queries/get_dd_by_id_with_rels.go
```diff
@@ -62,6 +64,19 @@ func (ch GetDeviceDefinitionWithRelsQueryHandler) Handle(ctx context.Context, qu
...
+			if !di.R.Integration.ManufacturerTokenID.IsZero() {
+				if m, ok := di.R.Integration.ManufacturerTokenID.Int64(); ok {
```
elffjs, line 69: I'm pretty sure there's a `Uint64` method.

internal/core/queries/get_dd_by_id_with_rels.go
```diff
@@ -62,6 +64,19 @@ func (ch GetDeviceDefinitionWithRelsQueryHandler) Handle(ctx context.Context, qu
...
+					manufTknID := uint64(m)
+					integrations[len(integrations)-1].ManufacturerTokenID = &manufTknID
```
elffjs, line 71: This read strangely to me. I think instead of

```
for _, a := range v {
	w = append(w, f(a))
	w[len(w)-1].x = g(a)
}
```

it might read better as

```
for _, a := range v {
	b := f(a)
	b.x = g(a)
	w = append(w, b)
}
```

elffjs (changes requested):

elffjs:
@JamesReate Asking for your crew's guidance here. Trying to fit in.

The motivation here is that we want to go from (token) pairing information to the `source = "dimo/integration/27qftVRWQYpVDcO5DltO5Ojbjxk"` like you see in Elastic. It's also an opportunity to stop hardcoding these points values.

internal/api/grpc_definitions_service.go
```diff
@@ -155,14 +155,19 @@ func (s *GrpcDefinitionsService) GetIntegrations(ctx context.Context, _ *emptypb
 				ICE:  int32(item.AutoPiPowertrainToTemplateID[coremodels.ICE]),
 				PHEV: int32(item.AutoPiPowertrainToTemplateID[coremodels.PHEV]),
 			},
+			Points: int64(item.Points),
 		})
+
+		if item.ManufacturerTokenID != nil {
+			result.Integrations[len(result.Integrations)-1].ManufacturerTokenId = item.ManufacturerTokenID.Uint64()
```
elffjs, line 162: Same criticism here.

JamesReate: yeah agree this would be good to change for more readable version

pkg/grpc/device_definition.proto
```diff
@@ -151,6 +151,8 @@ message Integration {
     int32 ICE = 3;
     int32 PHEV = 4;
   }
+  int64 points = 9;
+  uint64 manufacturer_token_id = 10;
```
JamesReate, line 155: don't know if we care about this being a uint64 here but big.Int elsewhere. my guess is it'll be fine b/c we won't ever have millions of different manufacturers. 

https://www.reddit.com/r/golang/comments/awtxw8/whats_the_best_way_to_add_a_bigint_to_a_protobuf/

JamesReate (approved):
just make same change pointed out by Dylan for readability and looks good.

pkg/grpc/device_definition.proto
elffjs: Oh there are Protobuf extensions:

https://github.com/nutsfinance/solidity-protobuf

internal/core/queries/get_in_all.go
```diff
@@ -54,7 +56,13 @@ func (ch GetAllIntegrationQueryHandler) Handle(ctx context.Context, _ mediator.M
 			AutoPiDefaultTemplateID: im.AutoPiDefaultTemplateID,
 			RefreshLimitSecs:        v.RefreshLimitSecs,
 			TokenID:                 v.TokenID.Int,
+			Points:                  v.Points,
+		}
+
+		if !v.ManufacturerTokenID.IsZero() {
+			result[i].ManufacturerTokenID = v.ManufacturerTokenID.Big.Int(new(big.Int))
```
elffjs, line 63: You can just do 
```suggestion
			result[i].ManufacturerTokenID = v.ManufacturerTokenID.Big.Int(nil)
```
I'm pretty sure.

internal/api/grpc_definitions_service.go
```diff
@@ -155,14 +155,20 @@ func (s *GrpcDefinitionsService) GetIntegrations(ctx context.Context, _ *emptypb
 				ICE:  int32(item.AutoPiPowertrainToTemplateID[coremodels.ICE]),
 				PHEV: int32(item.AutoPiPowertrainToTemplateID[coremodels.PHEV]),
 			},
-		})
+			Points: int64(item.Points),
+		}
+
+		if item.ManufacturerTokenID != nil {
```
elffjs, line 161: Appreciate it.

internal/core/models/integration_models.go
```diff
@@ -50,4 +51,6 @@ type GetIntegrationQueryResult struct {
 	RefreshLimitSecs             int                    `json:"refresh_limit_secs"`
 	AutoPiPowertrainToTemplateID map[PowertrainType]int `json:"auto_pi_power_train_to_template_id,omitempty"`
 	TokenID                      int                    `json:"token_id"`
+	Points                       int                    `json:"points"`
+	ManufacturerTokenID          *big.Int               `json:"manufacturer_token_id"`
```
elffjs, line 55: But it's weird that we use `*big.Int` for this and `int` for `TokenID`.

elffjs (approved):
Some minor concerns, but overall it's great.

internal/core/models/integration_models.go
JamesReate: I'd be fine changing it for consistency. But part of me also wonders if we just use uint64 and save the hassle of big.Int.

JamesReate:
> @JamesReate Asking for your crew's guidance here. Trying to fit in.
> 
> The motivation here is that we want to go from (token) pairing information to the `source = "dimo/integration/27qftVRWQYpVDcO5DltO5Ojbjxk"` like you see in Elastic. It's also an opportunity to stop hardcoding these points values.

yeah this looks good, happy to see the progress evolving it.

internal/api/grpc_definitions_service.go
```diff
@@ -176,7 +182,12 @@ func (s *GrpcDefinitionsService) prepareIntegrationResponse(integration coremode
 			ICE:  int32(integration.AutoPiPowertrainToTemplateID[coremodels.ICE]),
 			PHEV: int32(integration.AutoPiPowertrainToTemplateID[coremodels.PHEV]),
 		},
-	}, nil
+		Points: int64(integration.Points),
+	}
+	if integration.ManufacturerTokenID != 0 {
+		integ.ManufacturerTokenId = uint64(integration.TokenID)
+	}
```
elffjs, line 189: Same question here. Actually, is this a bug? We're assigning the integration's token id the "manufacturer token id" field?

internal/core/models/integration_models.go
```diff
@@ -50,4 +50,6 @@ type GetIntegrationQueryResult struct {
 	RefreshLimitSecs             int                    `json:"refresh_limit_secs"`
 	AutoPiPowertrainToTemplateID map[PowertrainType]int `json:"auto_pi_power_train_to_template_id,omitempty"`
 	TokenID                      int                    `json:"token_id"`
+	Points                       int                    `json:"points"`
+	ManufacturerTokenID          int                    `json:"manufacturer_token_id"`
```
elffjs, line 54: Why are the types different between here and `device_definition_models.go`?

elffjs (changes requested):

elffjs:
Linter?

elffjs:
Oh, I don't think that has anything to do with you. One second.

internal/core/models/integration_models.go
elffjs: Yeah who knows. Lots of duplication.

PR #205, opened 2024-02-13
SM-575 - Self Service add VIN decodings

rhvivancoeffio:
# Proposed Changes


JamesReate (approved):
lgtm but update proto generator before merging

JamesReate (changes requested):
also, looks like tests for decode vin need to be fixed / updated

PR #206, opened 2024-02-28
SM-544 - DAT Group VIN decoding

rhvivancoeffio:
# Proposed Changes


internal/core/services/vin_decoding_service.go
```diff
@@ -264,3 +289,30 @@ func buildFromDD(vin string, info *repoModel.DeviceDefinition) *models.VINDecodi
...
+	dossier := info.Body.GetDataVehicleIdentificationByVinResponse.VXS.Dossier[0].Vehicle
+	v := &models.VINDecodingInfoData{
+		VIN:  dossier.VINResult.VINVehicle.VINumber.VinCode,
+		Year: int32(dossier.BuildYear),
+		//Year:       2022,
+		Make:       strings.TrimSpace(dossier.ManufacturerName),
+		Model:      strings.TrimSpace(dossier.BaseModelName),
+		Source:     models.DATGroupProvider,
+		ExternalID: dossier.VINResult.VINVehicle.VINumber.VinCode,
+		//StyleName:  info.GetStyle(),
```
JamesReate, line 308: we'll need to think of something for the style name

internal/core/services/vin_decoding_service.go
```diff
@@ -84,6 +88,15 @@ func (c vinDecodingService) GetVIN(ctx context.Context, vin string, dt *repoMode
 		if err != nil {
 			return nil, err
 		}
+	case models.DATGroupProvider:
```
JamesReate, line 91: looks good to me

JamesReate (commented):
so far looks good, we'll just need to figure out what we use for the style, and tests

internal/core/services/vin_decoding_service.go
rhvivancoeffio: What would be the fields to be able to map them?

PR #208, opened 2024-03-08
SM-612 - Device definitions api dd on chain

rhvivancoeffio:
# Proposed Changes


internal/infrastructure/gateways/device_definition_on_chain_service.go
```diff
@@ -0,0 +1,104 @@
...
+		Logger:   logger,
+	}
+}
+
+func (e *deviceDefinitionOnChainService) CreateOrUpdate(ctx context.Context, manufacturerID types.NullDecimal, dd models.DeviceDefinition) (*string, error) {
+	if len(e.Settings.EthereumNetwork) == 0 {
+		return nil, nil
+	}
+
+	if manufacturerID.IsZero() {
```
LorranSutter, line 41: Shouldn't we instead check if the manufacturer Id exists/is valid?

internal/infrastructure/gateways/device_definition_on_chain_service.go
```diff
@@ -0,0 +1,104 @@
...
+	auth.GasLimit = uint64(300000)
+	auth.GasPrice = gasPrice
+	auth.From = fromAddress
+
+	instance, err := contracts.NewRegistryTransactor(contractAddress, client)
+	if err != nil {
+		return nil, fmt.Errorf("failed create NewRegistryTransactor: %w", err)
+	}
+
+	deviceInputs := []contracts.DeviceDefinitionInput{
```
LorranSutter, line 89: Looks like we are just adding one DD at a time here. If so, I think we should use the single `insertDeviceDefinition` instead of the batch to save some gas.

internal/config/settings.go
```diff
@@ -33,6 +33,9 @@ type Settings struct {
 	VincarioAPISecret              string         `yaml:"VINCARIO_API_SECRET"`
 	AutoIsoAPIUid                  string         `yaml:"AUTO_ISO_API_UID"`
 	AutoIsoAPIKey                  string         `yaml:"AUTO_ISO_API_KEY"`
+	EthereumNetwork                string         `yaml:"ETHEREUM_NETWORK"`
```
LorranSutter, line 36: I wonder if it would be useful to have different variables for dev and prod.

JamesReate:
- let's re-review this based on recent learnings. 
- rebase with main (lmk if need help)
- objective is: logic that creates or updates on device-definitions also pushes through to tableland / on-chain. Either from vin-decode or manual operations on Admin tool.

JamesReate:
closed since this is in prod.

PR #212, opened 2024-03-21
SM-650 - Device definitions api dd query tableland

rhvivancoeffio:
# Proposed Changes


charts/device-definitions-api/templates/secret.yaml
```diff
@@ -5,6 +5,18 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   data:
+  - remoteRef:
+      key: { { .Release.Namespace } }/definitions/ethereum/rpc_url
```
JamesReate, line 9: fix double spaces `{ {` should be `{{`

JamesReate (commented):
in readme, add an example command how to generate from ABI

JamesReate (approved):

PR #225, opened 2024-05-12
Resolve Image

rhvivancoeffio:
# Proposed Changes


JamesReate (commented):
made some tweaks - I wanted to avoid including metadata properties that did not have a value.

PR #231, opened 2024-05-14
Validate comma

rhvivancoeffio:
# Proposed Changes


internal/core/common/utils.go
```diff
@@ -569,5 +569,6 @@ func BuildDeviceIntegrationFeatureAttribute(attributes []*models.UpdateDeviceInt
 }
 
 func DeviceDefinitionSlug(makeSlug, modelSlug string, year int16) string {
```
JamesReate, line 571: let's just make sure this is being used everywhere that matters.

JamesReate (approved):

PR #238, opened 2024-06-14
[SM-911] add more validations for DAT group vin decoding

Asfiroth:
# Proposed Changes

### DIFF
--- a/internal/infrastructure/gateways/datgroup_api_service.go
+++ b/internal/infrastructure/gateways/datgroup_api_service.go
@@ -160,15 +160,13 @@ func parseXML(logger *zerolog.Logger, datgroupRespXML, vin string) (*DATGroupInf
 	seriesEquipment := xmlquery.FindOne(vehicle, "//ns1:SeriesEquipment")
 
 	if seriesEquipment == nil {
-		logger.Err(err).Str("vin", vin).Msgf("failed to find series equipment node")
-		return response, nil
+		return nil, errors.Wrap(err, "failed to find series equipment node")
 	}
 
 	seNodes := xmlquery.Find(seriesEquipment, "//ns1:EquipmentPosition")
 
 	if seNodes == nil {
-		logger.Err(err).Str("vin", vin).Msgf("failed to find series equipment nodes")
-		return response, nil
+		return nil, errors.Wrap(err, "failed to find series equipment position nodes")
 	}
 
 	for _, seNode := range seNodes {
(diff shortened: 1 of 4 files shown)


internal/infrastructure/gateways/datgroup_api_service.go
```diff
@@ -140,21 +140,37 @@ func parseXML(logger *zerolog.Logger, datgroupRespXML, vin string) (*DATGroupInf
...
 	} else {
 		response.Year = response.YearHigh
 	}
+
 	// series equipment
 	seriesEquipment := xmlquery.FindOne(vehicle, "//ns1:SeriesEquipment")
+
+	if seriesEquipment == nil {
+		logger.Err(err).Str("vin", vin).Msgf("failed to find series equipment node")
+		return response, nil
```
JamesReate, line 164: return nil, nil

internal/infrastructure/gateways/datgroup_api_service.go
```diff
@@ -140,21 +140,37 @@ func parseXML(logger *zerolog.Logger, datgroupRespXML, vin string) (*DATGroupInf
...
+	}
+
 	seNodes := xmlquery.Find(seriesEquipment, "//ns1:EquipmentPosition")
+
+	if seNodes == nil {
+		logger.Err(err).Str("vin", vin).Msgf("failed to find series equipment nodes")
+		return response, nil
```
JamesReate, line 171: return nil, err

Asfiroth: so it should always be equipments?

internal/infrastructure/gateways/datgroup_api_service.go
```diff
@@ -140,21 +140,37 @@ func parseXML(logger *zerolog.Logger, datgroupRespXML, vin string) (*DATGroupInf
...
 		response.Year = yr
 	} else {
 		response.Year = response.YearHigh
 	}
+
 	// series equipment
 	seriesEquipment := xmlquery.FindOne(vehicle, "//ns1:SeriesEquipment")
+
+	if seriesEquipment == nil {
+		logger.Err(err).Str("vin", vin).Msgf("failed to find series equipment node")
```
JamesReate, line 163: the err here would be from a different error, lets instantiate a new fmt.errorf(..) and then return that as well as log - although check the the caller already logs the error (may not be necessary to log just return nil, the error)

internal/infrastructure/gateways/datgroup_api_service.go
```diff
@@ -164,9 +180,22 @@ func parseXML(logger *zerolog.Logger, datgroupRespXML, vin string) (*DATGroupInf
 		}
 		response.SeriesEquipment = append(response.SeriesEquipment, equipment)
 	}
+
 	// special equipment
 	specialEquipment := xmlquery.FindOne(vehicle, "//ns1:SpecialEquipment")
+
+	if specialEquipment == nil {
+		logger.Err(err).Str("vin", vin).Msgf("failed to find special equipment node")
```
JamesReate, line 188: same here

JamesReate (approved):

PR #240, opened 2024-06-26
Adds NameSlug to decode vin response

KevinJoiner:
Retrieves NameSlug by joining with the device definition table.


JamesReate (approved):

internal/core/queries/decode_vin.go
```diff
@@ -106,8 +106,10 @@ func (dc DecodeVINQueryHandler) Handle(ctx context.Context, query mediator.Messa
 		return nil, errors.Wrap(err, "error when beginning transaction")
 	}
 	defer txVinNumbers.Rollback() //nolint
-
-	vinDecodeNumber, err := models.FindVinNumber(ctx, txVinNumbers, vin.String())
+	vinDecodeNumber, err := models.VinNumbers(
+		models.VinNumberWhere.Vin.EQ(vin.String()),
+		qm.Load(models.VinNumberRels.DeviceDefinition)).
+		One(ctx, dc.dbs().Reader)
```
elffjs, line 112: Should we keep using `txVinNumbers` here?

pkg/grpc/decoder.proto
```diff
@@ -19,6 +19,7 @@ message DecodeVinResponse {
   int32 year = 4;
   string source = 5;
   string powertrain = 6;
+  string name_slug =7;
```
elffjs, line 22:
```suggestion
  string name_slug = 7;
```

internal/core/queries/decode_vin.go
KevinJoiner: I was wondering if we need a transaction at all here since we are checking for existence and creating if it doesn't exist. Since there is nothing to rollback if a create fails.

KevinJoiner: It doesn't hurt so I will just use it, but I am curious if you can think of an edge case I am missing.

elffjs:
Test failure just looks like Docker stuff. Rerunning.

PR #245, opened 2024-07-11
[SM-970] add name slug (tableland id) to create dd response

Asfiroth:
# Proposed Changes


pkg/grpc/device_definition.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0
-// 	protoc        v4.25.2
+// 	protoc-gen-go v1.28.1
```
JamesReate, line 3: update your proto gen version

pkg/grpc/device_definition.proto
```diff
@@ -210,6 +210,11 @@ message BaseResponse {
   string id = 1;
 }
 
+message CreateDeviceDefinitionResponse {
```
JamesReate, line 213: is this only used by admin? i think so yeah?

PR #312, opened 2026-04-17
allow manual onboarding when VIN decode is low-confidence

zer0stars:
## Summary

Stops silent on-chain creation of garbage device definitions from low-confidence VIN decoders and hardens the Japan 17vin adapter.

### Background
Japanese engineers onboarding vehicles hit two failures:
- `GWS214-6014148` (Toyota Crown chassis) decoded via japan17vin as `Model="4D"` (body-style code), then got auto-registered on-chain as `toyota_4d_2017`. Client subsequently crashed trying to render the bad DD.
- Root cause: `japan_17vin_api.go` extracted model from a column that sometimes holds body-style codes, and `decode_vin.go` unconditionally created an on-chain DD on tableland miss regardless of how shaky the decode was.

### Changes
1. **`internal/infrastructure/gateways/japan_17vin_api.go`** — robust model extraction:
   - Tries `Model Name`, `Model name`, `车型` in priority order.
   - Rejects body-style codes matching `^\d+[A-Za-z]{0,3}$` (4D, 5D, 2D, 5HB).
   - Falls through to first non-body-style token in `Additional Vehicle Infomation`.
   - New unit tests.

2. **`internal/core/queries/decode_vin.go`** — gate auto-create for low-confidence sources:
   - If source is `japan17vin`, `carvxvin`, `autoiso`, or `eleva` and tableland returns no match → return `NotFoundError` instead of calling `Create()`.
   - Client (separate PR) catches this and opens the manual make/model/year picker.
   - High-confidence sources (Drivly, Vincario, DATGroup, Tesla) keep existing auto-create.

### Out of scope
- Cleaning up existing bad on-chain DDs (e.g., `toyota_4d_2017`).
- JPN 11-char chassis 500 for `GJ5FW-550049` (separate issue in shared `pkg/vin.IsValidJapanChassis()`).

## Test plan
- [x] `go test -race ./internal/infrastructure/gateways/...` — green, includes new japan17vin extraction tests.
- [x] `go build ./...` — clean.
- [x] `golangci-lint run` — no new issues.
- [ ] Local decode of `GWS214-6014148` with `countryCode=JPN` returns error (not a new on-chain DD).
- [ ] Local decode of a high-confidence VIN with tableland miss still auto-creates (regression check).
- [ ] End-to-end with dimo-driver companion PR: app opens picker on decode failure.

### DIFF
--- a/internal/core/queries/decode_vin.go
+++ b/internal/core/queries/decode_vin.go
@@ -246,9 +246,16 @@ func (dc DecodeVINQueryHandler) Handle(ctx context.Context, query *DecodeVINQuer
 		resp.Powertrain = pt
 	}
 
-	// if dd not found in tableland, we want to create it
+	// if dd not found in tableland, we want to create it — but only for high-confidence
+	// provider sources. Low-confidence sources (Japan chassis decoders, etc.) have produced
+	// garbage on-chain DDs (e.g. Model="4D" body-style codes), so we reject here and let
+	// the client fall back to manual make/model/year selection via the picker.
 	if tblDef != nil {
 		resp.DefinitionId = tblDef.ID
+	} else if isLowConfidenceSource(vinInfo.Source) {
+		metrics.InternalError.With(prometheus.Labels{"method": VinErrors}).Inc()
+		localLog.Warn().Str("decode_source", string(vinInfo.Source)).Msgf("low-confidence decode + tableland miss; returning not found so client opens manual picker")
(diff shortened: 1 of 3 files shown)


JamesReate (commented):
overall makes sense to me, doesn't seem like it should impact anything other than japan really. tests seem to cover a lot

zer0stars:
Folded into #314 so device-definitions-api has one open PR for this work. The commits are preserved there through merge commit 5694b45. The decode gate was ported onto #314's catalog create path: a catalog miss from japan17vin, carvxvin, autoiso or eleva answers not found instead of creating a template.

